PR #7, opened 2022-06-20
Allow a comment only once in a pull request

### LINKED ISSUE #4: Allow a comment only once in a pull request
Right now, it exists a `$comment()` action that can be executed if a rule is followed. 
However, this rule can be always followed whenever the workflow is triggered.
This can cause the same comment to be made several times in the same pull request which can be annoying at certain cases.

Develop an action that makes a comment only once in a pull request.

*None*.

**Additional context**
*None*.

shay2025:
This pull request introduces a new action called `$commentOnce()` that posts a comment in a pull request if a rule is followed and if the comment hasn't been made already.

It was added unit tests for this action and some mocks.

Closes #4


go.mod
```diff
@@ -13,11 +13,20 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
```
marcelosousa, line 16: Which version of go do you have? It seems that we have different version of go which is adding these require statements.

lang/aladino/parser.go
```diff
@@ -1,6 +1,7 @@
 // Code generated by goyacc -o lang/aladino/parser.go -p Aladino lang/aladino/parser.y. DO NOT EDIT.
 
-//line lang/aladino/parser.y:2
+//line lang/aladino/parser.y:1
```
marcelosousa, line 3: We need to understand why this generated file is different in your case.

plugins/aladino/actions.go
```diff
@@ -242,6 +242,41 @@ func commentCode(e aladino.Env, args []aladino.Value) error {
...
+
+	commentBody := args[0].(*aladino.StringValue).Val
+
+	comments, _, err := e.GetClient().Issues.ListComments(e.GetCtx(), owner, repo, prNum, &github.IssueListCommentsOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, comment := range comments {
+		commentAlreadyMade := *comment.Body == commentBody
```
marcelosousa, line 267: I think we should improve this check. Imagine that the comment is 'LGTM' and someone else put this comment. This would prevent this built-in to act. As we do with the comments in the report, we need to annotate the comment body so that it is more clear that this comment was made by Reviewpad.

plugins/aladino/actions.go
```diff
@@ -242,6 +242,41 @@ func commentCode(e aladino.Env, args []aladino.Value) error {
...
+func commentOnceCode(e aladino.Env, args []aladino.Value) error {
+	pullRequest := e.GetPullRequest()
+
+	prNum := utils.GetPullRequestNumber(pullRequest)
+	owner := utils.GetPullRequestOwnerName(pullRequest)
+	repo := utils.GetPullRequestRepoName(pullRequest)
+
+	commentBody := args[0].(*aladino.StringValue).Val
+
+	comments, _, err := e.GetClient().Issues.ListComments(e.GetCtx(), owner, repo, prNum, &github.IssueListCommentsOptions{})
```
marcelosousa, line 261: I think this request is not paginated. We need to improve this. Check for example the function `getPullRequestFiles`.

plugins/aladino/actions_test.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Fatalf("mockEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{aladino.BuildStringValue("Dummy Comment")}
+
+	err = commentOnceCode(*testEvalEnv, args)
+
+	assert.Nil(t, err)
+	assert.Equal(t, "Dummy Comment", commentCreated)
+}
```
marcelosousa, line 106: Please add a new line to the end of the file.

plugins/aladino/mocks.go
```diff
@@ -0,0 +1,119 @@
...
+		ctx,
+		client,
+		nil,
+		nil,
+		pr,
+		PluginBuiltIns(),
+	)
+
+	return &env, err
+}
```
marcelosousa, line 119: Same here.

plugins/aladino/mocks.go
```diff
@@ -0,0 +1,119 @@
+// Copyright (C) 2019-2022 Explore.dev - All Rights Reserved
+// Unauthorized copying of this file, via any medium is strictly prohibited
+// Proprietary and confidential
```
marcelosousa, line 3: This LICENSE is not correct.

plugins/aladino/actions_test.go
```diff
@@ -0,0 +1,106 @@
+// Copyright (C) 2019-2022 Explore.dev - All Rights Reserved
+// Unauthorized copying of this file, via any medium is strictly prohibited
+// Proprietary and confidential
```
marcelosousa, line 3: This LICENSE is not correct. We need to update your VSCode settings :-)

marcelosousa (changes requested):
Some preliminary comments. The comments tagged with the 🚀 are deal breakers to the merge. 
**Pair review with @ferreiratiago**

plugins/aladino/actions.go
shay2025: This request is also made in the function `FindReportComment`. Should it also be paginated?

plugins/aladino/mocks.go
shay2025: There are many files around the repository that don't have the correct licence. We need to handle this.

plugins/aladino/mocks.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return mock.NewMockedHTTPClient(mocks...)
+}
+
+func mockEnv(clientOptions ...mock.MockBackendOption) (*aladino.Env, error) {
+	client := github.NewClient(mockClient(clientOptions...))
+
+	owner := "foobar"
+	repo := "mocked-proj-1"
+	prNum := 6
```
marcelosousa, line 100: It is confusing that this information is different than the one in L22.

plugins/aladino/actions_test.go
```diff
@@ -0,0 +1,106 @@
...
+			}),
+		),
+	)
+	if err != nil {
+		log.Fatalf("mockEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{aladino.BuildStringValue("<!--@annotation-reviewpad-->Lorem Ipsum")}
+
+	err = commentOnceCode(*testEvalEnv, args)
```
marcelosousa, line 67:
```suggestion
	err = commentOnceCode(testEvalEnv, args)
```

plugins/aladino/actions_test.go
```diff
@@ -0,0 +1,106 @@
...
+			}),
+		),
+	)
+	if err != nil {
+		log.Fatalf("mockEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{aladino.BuildStringValue("<!--@annotation-reviewpad-->Dummy Comment")}
+
+	err = commentOnceCode(*testEvalEnv, args)
```
marcelosousa, line 102:
```suggestion
	err = commentOnceCode(testEvalEnv, args)
```

marcelosousa (changes requested):
Review of the tests.

plugins/aladino/mocks.go
shay2025: L22 refers to the pull request ID whereas this information refers to the pull request number which are different things.

ferreiratiago (changes requested):
First iteration on the pull request. Still missing a review on the test files.

ferreiratiago (approved):
LGTM

plugins/aladino/actions.go
```diff
@@ -7,12 +7,15 @@ package plugins_aladino
...
 	"fmt"
 	"log"
+	"regexp"
 
 	"github.com/google/go-github/v42/github"
 	"github.com/reviewpad/reviewpad/v2/lang/aladino"
 	"github.com/reviewpad/reviewpad/v2/utils"
 )
 
+const reviewpadCommentAnnotation = "<!--@annotation-reviewpad-->"
```
marcelosousa, line 17: This constant already exists in the report.go file. I think we should reuse it

plugins/aladino/actions.go
```diff
@@ -242,6 +245,45 @@ func commentCode(e aladino.Env, args []aladino.Value) error {
...
+
+	comments, err := utils.GetPullRequestComments(e.GetCtx(), e.GetClient(), owner, repo, prNum)
+	if err != nil {
+		return err
+	}
+
+	reviewpadCommentRegex := regexp.MustCompile(fmt.Sprintf("^%v", reviewpadCommentAnnotation))
+
+	for _, comment := range comments {
+		isReviewpadComment := reviewpadCommentRegex.Match([]byte(*comment.Body))
```
marcelosousa, line 273: I still think the best solution were would be to hash the contents of the comments instead of using regex Match. What do you think?

plugins/aladino/mocks.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/reviewpad/reviewpad/v2/lang/aladino"
+)
+
+const defaultMockPrNum = 123
+const defaultMockPrOwner = "foobar"
+const defaultMockPrRepoName = "default-mock-repo"
+
+func getDefaultMockPullRequestDetails() *github.PullRequest {
+	prNum := defaultMockPrNum
+	prId := int64(prNum)
```
marcelosousa, line 24: According to @shay2025 , the pull request number is different than the pull request id which makes sense. I think it is okay to leave it as is but it could also be confusing. WDYT?

plugins/aladino/mocks.go
```diff
@@ -0,0 +1,144 @@
...
+	)
+
+	return env, err
+}
+
+// mockDefaultHttpClient mocks an HTTP client with default values and ready for Aladino.
+// Being ready for Aladino means that at least the request to build Aladino Env need to be mocked.
+// As for now, at least two github request need to be mocked in order to build Aladino Env, mainly:
+// - Get pull request details (i.e. /repos/{owner}/{repo}/pulls/{pull_number})
+// - Get pull request files (i.e. /repos/{owner}/{repo}/pulls/{pull_number})
```
marcelosousa, line 112: The request here is the same as L111. I think there's a mistake here.

utils/github.go
```diff
@@ -80,3 +81,32 @@ func ParseNumPages(resp *github.Response) int {
 
 	return ParseNumPagesFromLink(link)
 }
+
+func GetPullRequestComments(ctx context.Context, client *github.Client, owner string, repo string, number int) ([]*github.IssueComment, error) {
```
marcelosousa, line 85: Since you have added this function, for consistency, please move the function `getPullRequestFiles` that lives in the file `lang/aladino/env.go` to here as well.

marcelosousa (changes requested):
Left some more comments! I think we are almost there 👍

ferreiratiago:
Closing in favor of #24

PR #24, opened 2022-06-22
Allow a comment only once in a pull request

ferreiratiago:
Development done by @shay2025 

This pull request introduces a new action called $commentOnce() that posts a comment in a pull request if a rule is followed and if the comment hasn't been made already.

It was added unit tests for this action and some mocks.

Closes https://github.com/reviewpad/reviewpad/issues/4


plugins/aladino/actions.go
```diff
@@ -242,6 +243,44 @@ func commentCode(e aladino.Env, args []aladino.Value) error {
...
+
+func commentOnceCode(e aladino.Env, args []aladino.Value) error {
+	pullRequest := e.GetPullRequest()
+
+	prNum := utils.GetPullRequestNumber(pullRequest)
+	owner := utils.GetPullRequestOwnerName(pullRequest)
+	repo := utils.GetPullRequestRepoName(pullRequest)
+
+	commentBody := args[0].(*aladino.StringValue).Val
+	commentBodyWithReviewpadAnnotation := fmt.Sprintf("\n%v\n%v", aladino.ReviewpadCommentAnnotation, commentBody)
```
shay2025, line 261: A strange behaviour was happening when we only had `"%v%v"`.
For some reason, when we had this, although the create issue comment request was giving a successful status code, the comment didn't appear in the github interface.
However, when it was added the `\n` at the beginning, before the reviewpad annotation, the comment now appeared in the interface.

shay2025: This was happening because the following was occuring:
1. A comment starting with "" was made.
2. When the action runs, we generate a new report
3. Before generating this new report, we find the old one and delete it (the functions `FindReportComment` and `DeleteReportComment` are executed in this order).
4. In the `FindReportComment` function, to find the report we were only using the regex `^<!--@annotation-reviewpad-->` when this regex not only finds the reports but also other comments made by reviewpad.
5. So once a comment with this condition was found it would be deleted right away and this was causing the non-appearance of the comment in the github interface since right after being created it was deleted.
6. To fix this, we specified even more the regex for the reviewpad report: `^<!--@annotation-reviewpad-->\\*\\*Reviewpad Report\\*\\*\n\n`. 
https://github.com/reviewpad/reviewpad/blob/908b598a366125e9b3b9c5dc033957df4dd55f5b/lang/aladino/report.go#L190-L192

marcelosousa (changes requested):
Please see the comment on the annotations. We are almost there 🚀

plugins/aladino/actions.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/reviewpad/reviewpad/v2/utils"
 )
 
+const ReviewpadCommentAnnotation = "<!--@annotation-reviewpad-->"
```
marcelosousa, line 17:
```suggestion
const ReviewpadCommentAnnotation = "<!--@annotation-reviewpad-single-comment-->"
```

marcelosousa (changes requested):

marcelosousa (approved):
Looks good! 👍

PR #25, opened 2022-06-23
Paginate list comments request in aladino report

### LINKED ISSUE #9: Paginate list comments request in aladino report
The function `FindReportComment` does a request to [list the pull request comments](https://github.com/reviewpad/reviewpad/blob/f39befdd97ef7c0ee2d4af1e8c227a94ffc1c371/lang/aladino/report.go#L183) which isn't paginated.

We should improve this request by paginating it. Look at `getPullRequestFiles` which does a paginated request to [list the pull request files](https://github.com/reviewpad/reviewpad/blob/f39befdd97ef7c0ee2d4af1e8c227a94ffc1c371/lang/aladino/env.go#L146).

*None*

**Additional context**
*None*

shay2025:
This pull request includes the following changes:
- Paginate list comments request made in `FindReportComment` function present in `lang/aladino/report.go`
- Added the `opts` argument in `GetPullRequestComments` function present in `utils/github.go`. This was added because when we want to make a request to list the comments of pull request we may want to add a set of options for this listing and this options may vary according to the context of the `GetPullRequestComments` function call. For example, the call made in `FindReportComment` in `lang/aladino/report.go` and the call made in `plugins/aladino/actions.go` differ.

Closes #9

### DIFF
--- a/plugins/aladino/actions.go
+++ b/plugins/aladino/actions.go
@@ -295,7 +295,7 @@ func commentOnceCode(e aladino.Env, args []aladino.Value) error {
 	commentBodyWithReviewpadAnnotation := fmt.Sprintf("%v%v", ReviewpadCommentAnnotation, commentBody)
 	commentBodyWithReviewpadAnnotationHash := sha256.Sum256([]byte(commentBodyWithReviewpadAnnotation))
 
-	comments, err := utils.GetPullRequestComments(e.GetCtx(), e.GetClient(), owner, repo, prNum)
+	comments, err := utils.GetPullRequestComments(e.GetCtx(), e.GetClient(), owner, repo, prNum, &github.IssueListCommentsOptions{})
 	if err != nil {
 		return err
 	}
--- a/lang/aladino/report.go
+++ b/lang/aladino/report.go
@@ -180,7 +180,7 @@ func FindReportComment(env Env) (*github.IssueComment, error) {
 	repo := utils.GetPullRequestRepoName(pullRequest)
 	prNum := utils.GetPullRequestNumber(pullRequest)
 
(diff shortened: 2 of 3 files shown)


plugins/aladino/actions.go
```diff
@@ -263,7 +263,7 @@ func commentOnceCode(e aladino.Env, args []aladino.Value) error {
 	commentBodyWithReviewpadAnnotation := fmt.Sprintf("%v%v", ReviewpadCommentAnnotation, commentBody)
 	commentBodyWithReviewpadAnnotationHash := sha256.Sum256([]byte(commentBodyWithReviewpadAnnotation))
 
-	comments, err := utils.GetPullRequestComments(e.GetCtx(), e.GetClient(), owner, repo, prNum)
+	comments, err := utils.GetPullRequestComments(e.GetCtx(), e.GetClient(), owner, repo, prNum, &github.IssueListCommentsOptions{})
```
marcelosousa, line 266: We need to add a unit test to ensure that this argument does not have unintended consequences.

marcelosousa (changes requested):
Please check the comment - thank you!

plugins/aladino/actions.go
shay2025: We can't test the effect of this specific argument with unit tests.
To do a unit test we need to mock a response.
Whatever the parameters we put in this `github.IssueListCommentsOptions{}` the response will always be the same since the operations applied to the list of comments (sort, etc) are done by github and so since we are mocking the response, this will always be the same showing that the supposed operations won't have any effect.

ferreiratiago (approved):
LGTM

marcelosousa (approved):

PR #30, opened 2022-06-27
Add YAML property to define failure mode

### LINKED ISSUE #28: YAML property to define failure mode
Although the engine runners return an error, this error is at the moment ignored by the runner users such as the [Reviewpad Action](https://github.com/reviewpad/action). It should be up to the user to specify if that is the intended behaviour.

Introduce a new boolean property in the YAML called "fail-on-error" which is `true` by default.

None.

**Additional context**
None.

shay2025:
This pull request adds a new YAML property to define failure mode called "skip-fail-on-error" which by default is `false`.

Closes #28.

### DIFF
--- a/engine/lang.go
+++ b/engine/lang.go
@@ -141,15 +141,15 @@ func (p PadGroup) equals(o PadGroup) bool {
 }
 
 type ReviewpadFile struct {
-	Version         string              `yaml:"api-version"`
-	Edition         string              `yaml:"edition"`
-	Mode            string              `yaml:"mode"`
-	SkipFailOnError bool                `yaml:"skip-fail-on-error"`
-	Imports         []PadImport         `yaml:"imports"`
-	Groups          []PadGroup          `yaml:"groups"`
-	Rules           []PadRule           `yaml:"rules"`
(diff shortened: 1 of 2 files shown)


engine/lang.go
```diff
@@ -141,14 +141,15 @@ func (p PadGroup) equals(o PadGroup) bool {
...
-	Mode      string              `yaml:"mode"`
-	Imports   []PadImport         `yaml:"imports"`
-	Groups    []PadGroup          `yaml:"groups"`
-	Rules     []PadRule           `yaml:"rules"`
-	Labels    map[string]PadLabel `yaml:"labels"`
-	Workflows []PadWorkflow       `yaml:"workflows"`
+	Version         string              `yaml:"api-version"`
+	Edition         string              `yaml:"edition"`
+	Mode            string              `yaml:"mode"`
+	SkipFailOnError bool                `yaml:"skip-fail-on-error"`
```
marcelosousa, line 147: Please rename this property to `ignore-errors`.

marcelosousa (changes requested):
Just a comment on the name of the property.

marcelosousa (approved):
LGTM

PR #42, opened 2022-07-01
Add unit tests for `$removeLabel` action builtin

shay2025:
This pull requests introduces the following:
- removal of args length and type check in removeLabel action builtin since these checks are already made before the action execution;
- add labels to the pull request mocks;
- add unit tests for the `removeLabel` action builtin.

### DIFF
--- a/plugins/aladino/actions/removeLabel_test.go
+++ b/plugins/aladino/actions/removeLabel_test.go
@@ -1,0 +1,103 @@
+// Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+// Use of this source code is governed by a license that can be
+// found in the LICENSE file.
+
+package plugins_aladino_actions_test
+
+import (
+	"log"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-github/v42/github"
+	"github.com/gorilla/mux"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	"github.com/reviewpad/reviewpad/v2/lang/aladino"
+	mocks_aladino "github.com/reviewpad/reviewpad/v2/mocks/aladino"
+	plugins_aladino "github.com/reviewpad/reviewpad/v2/plugins/aladino"
+	"github.com/stretchr/testify/assert"
+)
+
+var removeLabel = plugins_aladino.PluginBuiltIns().Actions["removeLabel"].Code
+
+func TestRemoveLabel_WhenGetLabelRequestFails(t *testing.T) {
+	failMessage := "GetLabelRequestFail"
+	mockedEnv, err := mocks_aladino.MockDefaultEnv(
+		mock.WithRequestMatchHandler(
+			mock.GetReposLabelsByOwnerByRepoByName,
+			http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				mock.WriteError(
+					w,
+					http.StatusInternalServerError,
+					failMessage,
+				)
+			}),
+		),
+	)
+	if err != nil {
+		log.Fatalf("mockDefaultEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{aladino.BuildStringValue("test")}
+	err = removeLabel(mockedEnv, args)
+
+	assert.Equal(t, err.(*github.ErrorResponse).Message, failMessage)
+}
+
+func TestRemoveLabel_WhenLabelIsNotAppliedToPullRequest(t *testing.T) {
+	label := "bug"
+	var labelRemoved bool
+	mockedEnv, err := mocks_aladino.MockDefaultEnv(
+		mock.WithRequestMatch(
+			mock.GetReposLabelsByOwnerByRepoByName,
+			github.Label{
+				Name: github.String("bug"),
+			},
+		),
+		mock.WithRequestMatchHandler(
+			mock.DeleteReposIssuesLabelsByOwnerByRepoByIssueNumberByName,
+			http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				// If the remove label request was performed then the label was removed
+				labelRemoved = true
+			}),
+		),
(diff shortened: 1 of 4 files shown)


plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+		),
+	)
+	if err != nil {
+		log.Fatalf("mockDefaultEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{aladino.BuildStringValue("test")}
+	err = removeLabel(mockedEnv, args)
+
+	assert.Equal(t, err.(*github.ErrorResponse).Message, failMessage)
```
ferreiratiago, line 44: We should be using `EqualError` for these cases.

```go
assert.EqualError(t, err, failMessage)
```

plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestRemoveLabel_WhenLabelIsNotAppliedToPullRequest(t *testing.T) {
```
ferreiratiago, line 47: I'm a bit confused with this test. From the description it say `WhenLabelIsNotAppliedToPullRequest` but then we expect the label to be removed `labelRemoved = true` and then on line 63 the description says `The label should not be removed`. Can you please clarify this?

plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+	label := "bug"
+	var labelRemoved bool
+	mockedEnv, err := mocks_aladino.MockDefaultEnv(
+		mock.WithRequestMatch(
+			mock.GetReposLabelsByOwnerByRepoByName,
+			github.Label{
+				Name: github.String("bug"),
```
ferreiratiago, line 54: Shouldn't we use `label` instead of the literal `bug`?

```go
Name: github.String(label)
```

plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestRemoveLabel_WhenLabelIsAppliedToPullRequest(t *testing.T) {
+	label := "enhancement"
+	var gotLabel string
+	mockedEnv, err := mocks_aladino.MockDefaultEnv(
+		mock.WithRequestMatch(
+			mock.GetReposLabelsByOwnerByRepoByName,
+			github.Label{
+				Name: github.String("enhancement"),
```
ferreiratiago, line 83: [reviewpad-sig]: <>
[Link to Comment](https://beta.reviewpad.com/review/github.com/reviewpad/reviewpad/pull/42#comment-5477e452-f867-469e-b71f-d2845f9e5e37)

plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	args := []aladino.Value{aladino.BuildStringValue("test")}
+	err = removeLabel(mockedEnv, args)
+
+	assert.Equal(t, err.(*github.ErrorResponse).Message, failMessage)
+}
+
+func TestRemoveLabel_WhenLabelIsNotAppliedToPullRequest(t *testing.T) {
+	label := "bug"
+	var labelRemoved bool
```
ferreiratiago, line 49: I would suggest to rename this to `isLabelRemoved`.

plugins/aladino/actions/removeLabel_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	args := []aladino.Value{aladino.BuildStringValue(label)}
+	err = removeLabel(mockedEnv, args)
+
+	assert.Nil(t, err)
+	assert.False(t, labelRemoved, "The label should not be removed")
+}
+
+func TestRemoveLabel_WhenLabelIsAppliedToPullRequest(t *testing.T) {
+	label := "enhancement"
```
ferreiratiago, line 77: I would suggest to rename this to `wantLabel`.

ferreiratiago (changes requested):

[reviewpad-sig]: <>
[View on **Reviewpad**](https://beta.reviewpad.com/review/github.com/reviewpad/reviewpad/pull/42?codereviewid=7e6eea98-7c42-4538-9deb-2c4e9f465381)

plugins/aladino/actions/removeLabel_test.go
shay2025: @ferreiratiago this follows a similar logic as [`TestCommentOnce_WhenCommentAlreadyExists`](https://github.com/reviewpad/reviewpad/blob/817cb2d3b356ed6b6edda465dad6aac1c7f22ec3/plugins/aladino/actions/commentOnce_test.go#L52-L82)
We need to test this scenario:
https://github.com/reviewpad/reviewpad/blob/f5bf22ce41fff44666018898467991aaa5d9379e/plugins/aladino/actions/removeLabel.go#L39-L41
But the result of this scenario is only a return of `nil` error which I believe isn't enough since if a label is successfully removed it will also return a `nil` error.
So to counter this, I propose we use a flag `labelRemoved` which will be an indicator if the request to remove a label was executed and if so then the condition `labelIsAppliedToPullRequest` is `true` and what we want to cover won't be covered.

plugins/aladino/actions/removeLabel_test.go
shay2025: @ferreiratiago Yes I agree but when handling `github.ErrorResponse` it becomes more complicated, i.e, the readability of the handling isn't neat:
```
args := []aladino.Value{aladino.BuildStringValue("test")}
err = errors.New(removeLabel(mockedEnv, args).(*github.ErrorResponse).Message)

assert.EqualError(t, err, errorMessage)
```

ferreiratiago (approved):
LGTM

PR #45, opened 2022-07-04
Add unit tests for `$close` action builtin

shay2025:
This pull request adds unit tests for `close` action builtin.


plugins/aladino/actions/close_test.go
```diff
@@ -0,0 +1,73 @@
...
+		),
+	)
+	if err != nil {
+		log.Fatalf("mockDefaultEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{}
+	err = close(mockedEnv, args)
+
+	assert.Equal(t, err.(*github.ErrorResponse).Message, failMessage)
```
ferreiratiago, line 45: Please use `EqualError`.

plugins/aladino/actions/close_test.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func TestClose(t *testing.T) {
```
ferreiratiago, line 48: I'm not sure how this is working. This test is using the default values but the default values for the pull request (i.e. `MockDefaultEnv`) do not define the state, but somehow the `closed` value is there. Even though it works we need to make sure that the mocked github pull request has the values we need. Let's discuss this offline.

ferreiratiago (changes requested):
We need to make sure the mocks that the value we need. Let's discuss this offline.

[reviewpad-sig]: <>
[View on **Reviewpad**](https://beta.reviewpad.com/review/github.com/reviewpad/reviewpad/pull/45?codereviewid=791cafb9-0f95-4499-a5dd-12726a5d3012)

plugins/aladino/actions/close_test.go
ferreiratiago: As discussed the `close` builtin sets the pull request state to `closed` and posts it. Therefore the value of `state` is `closed`. Sorry for the confusion :)

ferreiratiago (approved):
LGTM

PR #47, opened 2022-07-04
Restructure functions builtins

shay2025:
This pull request proposes a new way of organizing the functions builtins.


go.mod
```diff
@@ -16,6 +15,7 @@ require (
 )
 
 require (
+	github.com/bmatcuk/doublestar v1.3.4
```
ferreiratiago, line 18: This dependency should be in the `require` above. This require is for `indirect` dependencies.

ferreiratiago (changes requested):
Regarding this change I would suggest to do the same as for `actions`, i.e. add the package `plugins_aladino_functions` inside the folder `functions` and then add all functions in there (instead of creating package for each category).

I understand the reason to have these different packages but I'm afraid it brings extra complexity with no need. 

WDYT @marcelosousa ?

[reviewpad-sig]: <>
[View on **Reviewpad**](https://beta.reviewpad.com/review/github.com/reviewpad/reviewpad/pull/47?codereviewid=46956ed3-b495-4c19-92de-4f035806bc70)

go.mod
marcelosousa: I agree with your comment @ferreiratiago !

PR #57, opened 2022-07-04
fix error handling

GildedHonour:


runner.go
```diff
@@ -72,9 +72,13 @@ func Run(
 		}
 	}
 
-	evalEnv.Collector.Collect("Completed Analysis", map[string]interface{}{
+	err = evalEnv.Collector.Collect("Completed Analysis", map[string]interface{}{
 		"pullRequestUrl": evalEnv.PullRequest.URL,
 	})
 
+	if err != nil {
```
ferreiratiago, line 79: In this particular case we don't `reviewpad` to fail if the `collector` fails. This is because the `collector` is not essencial for `reviewpad` to succeed. My suggestion is to `log` the `collector` error and processed.  Suggestion:

```go
if err != nil {
	log.Printf("error on collector due to %v", err.Error())
}
```

Does this makes sense?

ferreiratiago (changes requested):
Hi @GildedHonour 👋🏼 Thanks for your contribution 🙌🏼  Please have a look at the comments. Also, if you could add some description to the pr that would be great :) Please feel free to join our [discord community](https://reviewpad.com/discord).

ferreiratiago (approved):
LGTM 👍🏼 Thanks for the contribution @GildedHonour 🙌🏼

Issue #49, opened 2022-07-04
Error in assignment of random reviewer


marcelosousa:
**Describe the bug**

The following Reviewpad run (https://github.com/reviewpad/reviewpad/runs/7179773626?check_suite_focus=true) failed with: 
```
can't assign a random user because there is no users
```

**To Reproduce**
Use the `$assignRandomReviewer` built-in?

**Expected behavior**
No errors.

PR #70, opened 2022-07-06
Fix assignment of random reviewer and add unit tests to `$assignRandomReviewer` action

shay2025:
This pull requests implements the fix to reported bug in #49 and adds unit tests to the `$assignRandomReviewer` action builtin.


plugins/aladino/actions/assignRandomReviewer_test.go
```diff
@@ -0,0 +1,171 @@
...
+		mock.WithRequestMatch(
+			mock.GetReposPullsRequestedReviewersByOwnerByRepoByPullNumber,
+			github.Reviewers{},
+		),
+		mock.WithRequestMatch(
+			mock.GetReposAssigneesByOwnerByRepo,
+			[]*github.User{
+				// Author of the mocked pull request
+				{Login: github.String("john")},
+				{Login: github.String("mary")},
```
ferreiratiago, line 146: These values need to be mocked. We shouldn't rely on the default mocks. Please have a look at the previous comment. 

Also, these values should be variable at the top of the function, in order to use them on `assert`.

plugins/aladino/actions/assignRandomReviewer_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		log.Fatalf("mockDefaultEnv failed: %v", err)
+	}
+
+	args := []aladino.Value{}
+	err = assignRandomReviewer(mockedEnv, args)
+
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(selectedReviewer))
+	assert.Equal(t, "mary", selectedReviewer[0])
```
ferreiratiago, line 170: This should be `assert.ElementsMatch(t, []string{"mary"}, selectedReviewers)`.

You can read more about ElementsMatch [here](https://github.com/stretchr/testify/blob/v1.8.0/assert/assertion_forward.go#L79).

plugins/aladino/actions/assignRandomReviewer_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	args := []aladino.Value{}
+	err = assignRandomReviewer(mockedEnv, args)
+
+	assert.EqualError(t, err, "can't assign a random user because there is no users")
+}
+
+func TestAssignRandomReviewer(t *testing.T) {
+	selectedReviewer := []string{}
```
ferreiratiago, line 135: Please rename this to `selectedReviewers`.

ferreiratiago (changes requested):

plugins/aladino/actions/assignRandomReviewer_test.go
shay2025: @ferreiratiago We only have default mocks for the pull request and these values are related to the repo.
And in the mocks the only parameter we can access in the `Repo` is `AssigneesURL` which isn't properly what we want.
image

ferreiratiago (approved):
Just a minor fix.

PR #73, opened 2022-07-06
Add unit tests for `$assignees` function builtin

shay2025:
## Description

This pull request adds unit tests for `$assignees` function builtin.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` command.

## Checklist


marcelosousa:
@shay2025 there's a typo - `assignes` -> `assignees`

plugins/aladino/functions/assignees_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/reviewpad/reviewpad/v2/lang/aladino"
+	mocks_aladino "github.com/reviewpad/reviewpad/v2/mocks/aladino"
+	plugins_aladino "github.com/reviewpad/reviewpad/v2/plugins/aladino"
+	"github.com/stretchr/testify/assert"
+)
+
+var assignees = plugins_aladino.PluginBuiltIns().Functions["assignees"].Code
+
+func TestAssignees(t *testing.T) {
```
ferreiratiago, line 19: We need to improve this test in order not to rely on the `getDefaultMockPullRequestDetails`.

For instance, if we remove `Assignees: []*github.User{{Login: github.String("jane")}}` from the mock, this test will continue to pass but it's not performing a proper test.

We need to mock our tests based on `getDefaultMockPullRequestDetails` but always changes whatever we are testing to make sure what when the test fails is for the right reason.

ferreiratiago (changes requested):
The test needs to be refactored. Please have a look at the comment :)

reviewpad-team:
**Reviewpad Report**

ferreiratiago (approved):
LGTM

PR #81, opened 2022-07-07
Add unit tests for `$createdAt` function builtin

shay2025:
## Description

This pull request adds unit tests for the `$createdAt` function builtin.
It wasn't possible to develop a unit test for the case of when the `time.Parse` throws an error since the `createdAt` parameter returned by GitHub is assumed to be valid at all times and so the parse will be successful at all times supposedly.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` command.

## Checklist

### DIFF
--- a/plugins/aladino/functions/createdAt_test.go
+++ b/plugins/aladino/functions/createdAt_test.go
@@ -6,9 +6,12 @@ package plugins_aladino_functions_test
 
 import (
 	"log"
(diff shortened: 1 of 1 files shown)


reviewpad-team:
**Reviewpad Report**

ferreiratiago (commented):
LGTM

PR #84, opened 2022-07-07
Add unit tests for `$group` function builtin

shay2025:
## Description

This pull request adds unit tests for the `$group` function builtin.

## Related issue

*None*

## Type of change

Improvements (non-breaking change without functionality

## How was this tested?

Ran `task test` command.

## Checklist

### DIFF
--- a/plugins/aladino/functions/group_test.go
+++ b/plugins/aladino/functions/group_test.go
@@ -19,7 +19,7 @@ var group = plugins_aladino.PluginBuiltIns().Functions["group"].Code
 
 func TestGroup(t *testing.T) {
   groupName := "techLeads"
-  mockedEnv, err := mocks_aladino.MockDefaultEnv()
+  mockedEnv, err := mocks_aladino.MockDefaultEnv(nil, nil)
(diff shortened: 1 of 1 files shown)


reviewpad-team:
**Reviewpad Report**

plugins/aladino/functions/group_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestGroup_WhenGroupIsNonExisting(t *testing.T) {
+  groupName := "techLeads"
+  mockedEnv, err := mocks_aladino.MockDefaultEnv()
+	if err != nil {
+		log.Fatalf("mockDefaultEnv failed: %v", err)
+	}
+
+  // Make sure that the group 'techLeads' doesn't exist
+  delete(mockedEnv.GetRegisterMap(), "techLeads")
```
ferreiratiago, line 46: No need for this. Is this happens is because something is wrong with the test or the builtin implementation, which is that case we want it to fail.

ferreiratiago (approved):

PR #122, opened 2022-07-13
test: lang aladino builder unit tests

shay2025:
## Description

This pull request introduces the unit tests for aladino builder.
In order to develop this tests, we had to change the access to the testing functions since they would be needed to be exported.
This caused the `parser.y` to also be changed.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` command.

## Checklist

### DIFF
--- a/lang/aladino/expr.go
+++ b/lang/aladino/expr.go
@@ -245,7 +245,7 @@ type Variable struct {
 	ident string
 }
 
-func variable(ident string) *Variable {
+func VariableConstr(ident string) *Variable {
 	return &Variable{ident}
 }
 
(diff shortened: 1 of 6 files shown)


lang/aladino/expr.go
```diff
@@ -245,7 +245,7 @@ type Variable struct {
 	ident string
 }
 
-func variable(ident string) *Variable {
+func VariableConstr(ident string) *Variable {
```
ferreiratiago, line 248: Let's call this `VariableBuilder` and apply the same idea to the other functions.

ferreiratiago (changes requested):

marcelosousa (approved):

PR #125, opened 2022-07-14
test: nil file test case in $hasCodePattern unit tests

shay2025:
## Description

This pull request completes the unit tests for `$hasCodePattern` giving a coverage of 100% for this function builtin.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` command.

## Checklist

### DIFF
--- a/plugins/aladino/functions/hasCodePattern_test.go
+++ b/plugins/aladino/functions/hasCodePattern_test.go
@@ -20,9 +20,10 @@ import (
 var hasCodePattern = plugins_aladino.PluginBuiltIns().Functions["hasCodePattern"].Code
 
 func TestHasCodePattern_WhenPullRequestPatchHasNilFile(t *testing.T) {
+    fileName := "default-mock-repo/file1.ts"
 	mockedPullRequestFileList := &[]*github.CommitFile{{
 		Patch:    nil,
-		Filename: github.String("default-mock-repo/file1.ts"),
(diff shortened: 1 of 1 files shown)


reviewpad-team:
**Reviewpad Report**

plugins/aladino/functions/hasCodePattern_test.go
```diff
@@ -39,6 +40,8 @@ func TestHasCodePattern_WhenPullRequestPatchHasNilFile(t *testing.T) {
 		log.Fatalf("mockDefaultEnv failed: %v", err)
 	}
 
+    mockedEnv.GetPatch()[fileName] = nil
```
ferreiratiago, line 43: Does this makes sense since we are mocking the `Patch` on line [25](https://github.com/reviewpad/reviewpad/pull/125/files#diff-c7812721a05bcb154055c03fdb0ce63bd70d874b7dad35c9e5a65e374d50b1ffR25) as nil?

shay2025: Even if we mock it, instead of `nil` we'll get:
```
&aladino.File{
  Repr: &github.CommitFile{
	Filename: github.String(filename),
  },
  Diff: []*aladino.DiffBlock{},
}
```
which isn't `nil` so in order to test the case in question we need to enforce the `nil` value.

PR #131, opened 2022-07-14
test: lang aladino env unit tests

shay2025:
## Description

This pull request introduces unit tests for `lang/aladino/env`.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` cmd.

## Checklist

### DIFF
--- a/lang/aladino/env_test.go
+++ b/lang/aladino/env_test.go
@@ -36,7 +36,7 @@ func TestGetCollector_WithDefaultEnv(t *testing.T) {
 		log.Fatalf("mockDefaultEnv failed: %v", err)
 	}
 
-	wantCollector := aladino.DefaultCollector
+	wantCollector := aladino.DefaultMockCollector
 
 	gotCollector := mockedEnv.GetCollector()
 
@@ -123,6 +123,7 @@ func TestGetBuiltIns_WithDefaultEnv(t *testing.T) {
 	assert.Equal(t, len(wantBuiltIns.Functions), len(gotBuiltIns.Functions))
 	assert.Equal(t, len(wantBuiltIns.Actions), len(gotBuiltIns.Actions))
 
+	// TODO: Search a way to compare functions
 	for functionName, functionCode := range wantBuiltIns.Functions {
 		assert.NotNil(t, gotBuiltIns.Functions[functionName])
 		assert.Equal(t, functionCode.Type, gotBuiltIns.Functions[functionName].Type)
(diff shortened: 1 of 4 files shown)


ferreiratiago (changes requested):
Please review the use of `reflect.DeepEqual` to make sure it's the right function to compare the output.

lang/aladino/env_test.go
```diff
@@ -0,0 +1,250 @@
...
+		"emptyFunction": aladino.BuildFunctionType([]aladino.Type{}, nil),
+		"emptyAction":   aladino.BuildFunctionType([]aladino.Type{}, nil),
+	})
+
+	gotTypeEnv := *aladino.NewTypeEnv(mockedEnv)
+
+	assert.Equal(t, wantTypeEnv, gotTypeEnv)
+}
+
+// The MockDefaultEnv function to create a mocked default env, it uses the NewEvalEnv which is the function we wish to test
```
marcelosousa, line 174: This comment is a bit confusing. From the name of the function, we should be testing NewEval but it's only called indirectly through the MockDefaultEnv. It seems that what we are really testing is the MockDefaultEnv.

lang/aladino/env_test.go
```diff
@@ -0,0 +1,250 @@
...
+				}),
+			),
+		},
+		nil,
+	)
+
+	assert.Equal(t, err.(*github.ErrorResponse).Message, failMessage)
+}
+
+func TestNewEvalEnv_WhenNewFileFails(t *testing.T) {
```
marcelosousa, line 196: Same comment from previous function applies here and in the next test functions.

marcelosousa (changes requested):
Thanks for this! I've left comments in the diff.

marcelosousa (commented):
Please check the comments.

lang/aladino/mocks.go
```diff
@@ -15,19 +15,23 @@ import (
...
-const defaultMockPrID = 1234
-const defaultMockPrNum = 6
-const defaultMockPrOwner = "foobar"
-const defaultMockPrRepoName = "default-mock-repo"
+const DefaultMockPrID = 1234
+const DefaultMockPrNum = 6
+const DefaultMockPrOwner = "foobar"
+const DefaultMockPrRepoName = "default-mock-repo"
+
+var DefaultContext = context.Background()
```
marcelosousa, line 27: please rename to DefaultMockX

marcelosousa (approved):
🙌

PR #158, opened 2022-07-24
feat: :sparkles: comment dry run report

zolamk:
## Description

- send dry run report

## Related issue

reviewpad/action#10

## Type of change

## How was this tested?

1. Run Reviewpad in dry run mode checked the results were as expected

## Checklist

### DIFF
--- a/lang/aladino/interpreter.go
+++ b/lang/aladino/interpreter.go
@@ -114,10 +114,7 @@ func (i *Interpreter) EvalExpr(kind, expr string) (bool, error) {
 }
 
 func (i *Interpreter) ExecProgram(program *engine.Program) error {
-	safeMode := i.Env.GetSafeMode()
-	dryRun := i.Env.GetDryRun()
-
-	execLogf("executing program: safe mode: %v, dry run: %v", safeMode, dryRun)
+	execLog("executing program")
 
 	for _, statement := range program.Statements {
 		err := i.ExecStatement(statement)
@@ -126,14 +123,12 @@ func (i *Interpreter) ExecProgram(program *engine.Program) error {
 		}
 	}
 
-	execLogf("execution done: safe mode: %v, dry run: %v", safeMode, dryRun)
+	execLog("execution done")
 
 	return nil
 }
 
 func (i *Interpreter) ExecStatement(statement *engine.Statement) error {
-	safeMode := i.Env.GetSafeMode()
-	dryRun := i.Env.GetDryRun()
 	statRaw := statement.Code
 	statAST, err := Parse(statRaw)
 	if err != nil {
@@ -154,22 +149,42 @@ func (i *Interpreter) ExecStatement(statement *engine.Statement) error {
 
 	i.Env.GetReport().addToReport(statement)
 
-	execLogf("\taction %v executed: safe mode: %v, dry run: %v", statRaw, safeMode, dryRun)
+	execLogf("\taction %v executed", statRaw)
 	return nil
 }
 
-func (i *Interpreter) Report(mode string) error {
-
+func (i *Interpreter) Report(mode string, safeMode bool) error {
 	env := i.Env
 
-	return i.Env.GetReport().SendReport(env.GetCtx(), env.GetSafeMode(), mode, env.GetPullRequest(), env.GetClient())
+	execLog("generating report")
+
+	var err error
+
+	comment, err := FindReportComment(env)
+	if err != nil {
+		return err
+	}
+
+	if mode == engine.SILENT_MODE {
+		if comment != nil {
+			return DeleteReportComment(env, *comment.ID)
+		}
+		return nil
+	}
+
+	report := buildReport(safeMode, env.GetReport())
+
+	if comment == nil {
+		return AddReportComment(i.Env, report)
+	}
+
+	return UpdateReportComment(i.Env, *comment.ID, report)
 
 }
 
 func NewInterpreter(
 	ctx context.Context,
 	dryRun bool,
-	safeMode bool,
 	gitHubClient *github.Client,
 	gitHubClientGQL *githubv4.Client,
 	collector collector.Collector,
@@ -178,7 +193,7 @@ func NewInterpreter(
 	builtIns *BuiltIns,
 
 ) (engine.Interpreter, error) {
-	evalEnv, err := NewEvalEnv(ctx, dryRun, safeMode, gitHubClient, gitHubClientGQL, collector, pullRequest, eventPayload, builtIns)
+	evalEnv, err := NewEvalEnv(ctx, dryRun, gitHubClient, gitHubClientGQL, collector, pullRequest, eventPayload, builtIns)
 	if err != nil {
 		return nil, err
 	}
--- a/lang/aladino/interpreter.go
+++ b/lang/aladino/interpreter.go
@@ -154,9 +154,9 @@ func (i *Interpreter) ExecStatement(statement *engine.Statement) error {
 }
 
 func (i *Interpreter) Report(mode string, safeMode bool) error {
-	env := i.Env
-
 	execLog("generating report")
+
+	env := i.Env
 
 	var err error
 
@@ -175,10 +175,10 @@ func (i *Interpreter) Report(mode string, safeMode bool) error {
 	report := buildReport(safeMode, env.GetReport())
 
 	if comment == nil {
-		return AddReportComment(i.Env, report)
+		return AddReportComment(env, report)
 	}
 
-	return UpdateReportComment(i.Env, *comment.ID, report)
+	return UpdateReportComment(env, *comment.ID, report)
 
 }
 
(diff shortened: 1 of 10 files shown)


reviewpad-team:
**Reviewpad Report**

* @marcelosousa: you are being notified because critical code was modified

lang/aladino/report.go
```diff
@@ -65,21 +94,39 @@ func (report *Report) addToReport(statement *engine.Statement) {
...
+
+func ReportHeader(dryRun bool) string {
 	var sb strings.Builder
 
 	// Annotation
 	sb.WriteString(fmt.Sprintf("%v\n", ReviewpadReportCommentAnnotation))
 	// Header
-	sb.WriteString("**Reviewpad Report**\n\n")
+	if dryRun {
+		sb.WriteString("**Reviewpad Report** (Reviewpad ran in dry-run mode because reviewpad.yml has changed)\n\n")
```
marcelosousa, line 118: We can actually pass a different input file. Instead of reviewpad.yml lets say that "because configuration file has changed"

runner.go
```diff
@@ -71,6 +71,14 @@ func Run(
 			engine.CollectError(evalEnv, err)
 			return nil, err
 		}
+	} else {
```
marcelosousa, line 74: This can be problematic because we can run Reviewpad as a CLI in dry-run in which case the report will be added to the pull request. I think we need to introduce a new boolean flag to signal when the configuration file was changed.

marcelosousa (changes requested):
Many thanks for this contribution 🙌 ! Please check the comment in runner.go which involves a substantial change in this pull request.

runner.go
zolamk: should we also allow setting the new `reviewpadChanged` boolean from the cli?

zolamk: Also, should it run in dry-run mode if the `reviewpadChanged` bool is true, or should we let the caller of the run function determine that?

marcelosousa: I think the boolean variable should be called something on the lines `safeMode` instead of `reviewpadChanged`. 

I don't think there's need for the new boolean to be passed in the CLI for now. 

Regarding the second question, yes, it's a pre-condition that if the new boolean is true then dry-run needs to be true as well.

lang/aladino/report.go
```diff
@@ -42,6 +43,34 @@ func mergeReportWorkflowDetails(left, right ReportWorkflowDetails) ReportWorkflo
 	return left
 }
 
+func (r *Report) SendReport(ctx context.Context, reviewpadFileChanged bool, mode string, pr *github.PullRequest, client *github.Client) error {
+
```
marcelosousa, line 47: Please remove this empty line. Same in the rest of the pull request!

marcelosousa (changes requested):
Please check the style issue!

marcelosousa:
@zolamk can you please update with the base branch? The build is failing at the moment.

zolamk:
> @zolamk can you please update with the base branch? The build is failing at the moment.

I am rebasing at the moment

> @zolamk can you please update with the base branch? The build is failing at the moment.

I've rebased and updated the tests.

runner.go
```diff
@@ -43,7 +44,12 @@ func Run(
 	eventPayload interface{},
 	reviewpadFile *engine.ReviewpadFile,
 	dryRun bool,
+	safeMode bool,
 ) (*engine.Program, error) {
+	if safeMode && !dryRun {
+		return nil, fmt.Errorf("reviewpad can only be run in dry-run when in safe mode")
```
marcelosousa, line 50: The error message doesn't seem correct because you can run reviewpad with dry-run as true and safe mode as false. What you mean here is: "when reviewpad is running in safe mode, it must also run in dry-run".

marcelosousa (changes requested):
Check the comments in the runner.go. Thanks!

runner.go
```diff
@@ -73,6 +79,16 @@ func Run(
...
 
+	if safeMode {
+		report := aladino.ReportFromProgram(program)
+
+		err = report.SendReport(ctx, dryRun, reviewpadFile.Mode, pullRequest, client)
+		if err != nil {
+			engine.CollectError(evalEnv, err)
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
zolamk: That was the first thing I tried too, but from what I can tell the aladino report is generated when the program is executed not during evaluation, which means in dry run mode since the program isn't executed there is no report generated and calling the `Report` method on the interpreter would just say `No workflows activated` [https://github.com/reviewpad/reviewpad/blob/62b1be11caa894e5cf04a252d72eae3cc807f29f/lang/aladino/interpreter.go#L131-L152](https://github.com/reviewpad/reviewpad/blob/62b1be11caa894e5cf04a252d72eae3cc807f29f/lang/aladino/interpreter.go#L131-L152)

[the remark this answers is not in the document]
zolamk: What I see as a possible solution is during evaluation we could keep another report maybe called `EvaulationReport` and during `safeMode` we could have an `EvaluationReport` method on the interpreter and use that for the report. Sounds good?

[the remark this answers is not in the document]
marcelosousa: I think it's confusing to have multiple reports. I think it's more elegant to pass the control variables to the interpreter environment and to use those variables inside the ExecProgram and ExecStatement interface method implementations. Does that sound good?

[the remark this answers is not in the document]
zolamk: Yeah, that's a much better solution.

runner.go
```diff
@@ -59,18 +65,19 @@ func Run(
...
+		return nil, err
+	}
+
+	err = aladinoInterpreter.Report(reviewpadFile.Mode)
+	if err != nil {
+		engine.CollectError(evalEnv, err)
+		return nil, err
+	}
+
+	if safeMode {
```
marcelosousa, line 80: This conditional is a no-op.

lang/aladino/interpreter.go
```diff
@@ -140,47 +145,31 @@ func (i *Interpreter) ExecStatement(statement *engine.Statement) error {
...
-		}
-		return nil
-	}
-
-	report := buildReport(env.GetReport())
-
-	if comment == nil {
-		return AddReportComment(env, report)
-	}
+	return i.Env.GetReport().SendReport(env.GetCtx(), env.GetSafeMode(), mode, env.GetPullRequest(), env.GetClient())
```
marcelosousa, line 165: This change increases the complexity of the code quite a bit. I think that abstracting the code with the `SendReport` makes things harder. I think we should revert that change.

Also, I think this is the only place where you really need to retrieve the `safeMode` boolean so I would prefer if you passed it as an argument to the `Report` function instead of adding it to the environment.

runner.go
```diff
@@ -59,18 +65,19 @@ func Run(
...
-			engine.CollectError(evalEnv, err)
-			return nil, err
-		}
+	err = aladinoInterpreter.ExecProgram(program)
+	if err != nil {
+		engine.CollectError(evalEnv, err)
+		return nil, err
+	}
+
+	err = aladinoInterpreter.Report(reviewpadFile.Mode)
```
marcelosousa, line 74: There's an issue with simply calling Report here. If we are running in safe mode then we want to add the report to the pull request, but if are not running in safe mode **and** in dry run, the comment will be added to the pull request and that is not intended.

marcelosousa (changes requested):

runner.go
```diff
@@ -59,14 +65,16 @@ func Run(
...
-		}
+	err = aladinoInterpreter.ExecProgram(program)
+	if err != nil {
+		engine.CollectError(evalEnv, err)
+		return nil, err
+	}
 
-		err = aladinoInterpreter.Report(reviewpadFile.Mode)
+	// only send report if reviewpad is running in safe mode
+	// or not in dry run mode
```
marcelosousa, line 75: this comment is not necessary :-)

lang/aladino/interpreter.go
```diff
@@ -170,17 +172,19 @@ func (i *Interpreter) Report(mode string) error {
 		return nil
 	}
 
-	report := buildReport(env.GetReport())
+	report := buildReport(safeMode, env.GetReport())
 
 	if comment == nil {
-		return AddReportComment(env, report)
+		return AddReportComment(i.Env, report)
```
marcelosousa, line 178:
```suggestion
		return AddReportComment(env, report)
```

lang/aladino/interpreter.go
```diff
@@ -170,17 +172,19 @@ func (i *Interpreter) Report(mode string) error {
...
 	}
 
-	return UpdateReportComment(env, *comment.ID, report)
+	return UpdateReportComment(i.Env, *comment.ID, report)
```
marcelosousa, line 181:
```suggestion
	return UpdateReportComment(env, *comment.ID, report)
```

marcelosousa (commented):
I think this is the last round of comments before it is ready to merge! Thank you so much @zolamk 🚀

marcelosousa (approved):

marcelosousa:
Many thanks for this contribution @zolamk 🚀

PR #163, opened 2022-07-25
test: add unit tests for lang/aladino expr

shay2025:
## Description

This pull request introduces unit tests for `lang/aladino/expr.go` file.
The only cases that are not covered in the unit tests are the ones that return a `log.Fatal`.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` cmd.

## Checklist


lang/aladino/expr_internal_test.go
```diff
@@ -0,0 +1,721 @@
...
+	now := time.Now()
+
+	val := "1 year ago"
+
+	timeValueRegex := regexp.MustCompile(`^[0-9]+`)
+	timeValue, _ := strconv.Atoi(timeValueRegex.FindString(val))
+
+	wantVal := &IntConst{
+		value: int(now.AddDate(-timeValue, 0, 0).Unix()),
+	}
```
marcelosousa, line 250: IMO, this is not the correct way to test this function because you're simply copying the current implementation. Therefore you're not really testing the expected behavior.

I think the correct way to test it to replace these lines by:
```
	wantVal := &IntConst{
		value: int(now.AddDate(-1, 0, 0).Unix()),
	}
```

lang/aladino/expr_internal_test.go
```diff
@@ -0,0 +1,721 @@
...
+func TestBuildRelativeTimeConst_WhenTimeUnitIsMonth_ExpectIntConstWithAValue(t *testing.T) {
+	now := time.Now()
+
+	val := "1 month ago"
+
+	timeValueRegex := regexp.MustCompile(`^[0-9]+`)
+	timeValue, _ := strconv.Atoi(timeValueRegex.FindString(val))
+
+	wantVal := &IntConst{
+		value: int(now.AddDate(0, -timeValue, 0).Unix()),
```
marcelosousa, line 266: Same comment as above and applies to ones below.

marcelosousa (changes requested):
Some comments about the way you're testing the relative timestamps. 

We need to agree on the conventions for the unit tests naming, some of the unit tests in this file are unreadable. 

@ferreiratiago do you have any opinion about the unit tests naming?

marcelosousa (approved):
LGTM.

lang/aladino/expr_internal_test.go
```diff
@@ -0,0 +1,701 @@
...
+}
+
+func TestVariableEquals_WhenEqual(t *testing.T) {
+	variable := BuildVariable("test")
+	otherVal := BuildVariable("test")
+
+	assert.True(t, variable.equals(otherVal))
+}
+
+func TestBuildUnaryOp_(t *testing.T) {
```
ferreiratiago, line 368: Please remove the extra `_`

ferreiratiago (approved):
LGTM

PR #167, opened 2022-07-26
test: add unit tests for lang/aladino typeinfer

shay2025:
## Description

This pull request introduces unit tests for `lang/aladino` typeinfer.
It also removes the period from error messages because error strings should not end with punctuation since they are usually printed following other context (https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` cmd

## Checklist


marcelosousa (approved):
Pairing with @shay2025

PR #169, opened 2022-07-26
chore: remove generated files from code coverage

shay2025:
## Description

This pull request introduces a way to remove the generated files from code coverage.
For this, it is executed a bash script (`scripts/exclude-from-code-coverage.sh`) upon running the `task test` cmd which will remove from the `coverage.out` the filenames that match the file patterns contained in `exclude-from-code-coverage.txt`.

## Related issue

*None*

## Type of change

## How was this tested?

Ran the following commands:
1. `task test`
2. `go tool cover -func coverage.out` - this command shows the code coverage for every single package withing your project and down at the bottom the total coverage. Taking this into account, it was verified that there was no `lang/aladino/parser.go` generated file included in the code coverage.

## Checklist


scripts/exclude-from-code-coverage.sh
```diff
@@ -0,0 +1,6 @@
+# Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+# Use of this source code is governed by a license that can be
+# found in the LICENSE file.
+
+#!/bin/sh
+grep -vf exclude-from-code-coverage.txt coverage.out > coverage.out.tmp && mv coverage.out.tmp coverage.out
```
marcelosousa, line 6: Please add a new line here.

exclude-from-code-coverage.txt
```diff
@@ -0,0 +1 @@
+lang/aladino/parser.go
```
marcelosousa, line 1: Please move this file to a directory of `resources` (probably inside the `scripts`). I don't think we should pollute the root of the repository with these files.

Taskfile.yml
```diff
@@ -48,6 +48,7 @@ tasks:
       Run with --watch or -w to watch for changes on _test.go files.
     cmds:
       - gotestsum -- -coverprofile=coverage.out ./...
+      - . scripts/exclude-from-code-coverage.sh
```
marcelosousa, line 51: Just a question: why do we need the `.` here?

marcelosousa (changes requested):
Thanks 🙌  I left some comments in the diff!

Taskfile.yml
shay2025: If we run `./scripts/exclude-from-code-coverage.sh` as it is usually done, it is thrown an error in Windows OS `%1 is not a valid Win32 application`. This seems to fix the problem but further investigation is needed.

marcelosousa (approved):
LGTM!

PR #173, opened 2022-07-26
test: add unit tests for engine program

shay2025:
## Description

This pull request introduces unit tests for engine program.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` cmd.

## Checklist

### DIFF
--- a/engine/program_internal_test.go
+++ b/engine/program_internal_test.go
@@ -20,19 +20,15 @@ func TestAppend(t *testing.T) {
 		},
 		Actions: []string{"$actionA()"},
 	}
-	actions := []string{"$actionB()"}
-	rules := []PadWorkflowRule{
-		{Rule: "test-rule-B"},
-	}
 
 	statement := &Statement{
-		Code: "$actionC()",
+		Code: "$actionB()",
 		Metadata: &Metadata{
 			Workflow: PadWorkflow{
-				Name: "test-workflow-C",
+				Name: "test-workflow-B",
 			},
 			TriggeredBy: []PadWorkflowRule{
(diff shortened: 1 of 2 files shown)


engine/program_internal_test.go
```diff
@@ -0,0 +1,69 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	wantProgram := &Program{
+		Statements: []*Statement{
+			{
+				Code: "$actionC()",
```
marcelosousa, line 46: Since you are re-using the `workflow` and the `rules` variables, it makes sense to create a variable `code`, `metadata`, etc to use here and in L30, L31, etc.

marcelosousa (changes requested):

engine/program_internal_test.go
shay2025: @marcelosousa is it enough to instead only have a `statement` variable?
```
statement := &Statement{
    Code: "$actionC()",
    Metadata: &Metadata{
	    Workflow: PadWorkflow{
		    Name: "test-workflow-C",
	    },
	    TriggeredBy: []PadWorkflowRule{
		    {Rule: "test-rule-C"},
	    },
    },
}
program := &Program{
    Statements: []*Statement{statement},
}

wantProgram := &Program{
    Statements: []*Statement{
	    statement,
	    {
		    Code: "$actionB()",
		    Metadata: &Metadata{
			    Workflow:    workflow,
			    TriggeredBy: rules,
		    },
	    },
    },
}
```

marcelosousa (approved):
Pair reviewed with @shay2025 .

PR #174, opened 2022-07-27
test: add unit tests for engine loader

shay2025:
## Description

This pull request introduces unit tests for engine loader.
To be able to do http mocks, it's proposed the use of https://github.com/jarcoal/httpmock.

Besides this, this pull request proposes the following improvements:
- A `resources` file to store `.yml` examples for test purposes, removing noise in test files such as having bytes var representing files contents;
- Some tests variables renaming.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task test` cmd.

## Checklist


marcelosousa (changes requested):
Please move the `resources/test` files to `testdata` folder.

Please rebase with the base branch.

engine/loader_internal_test.go
```diff
@@ -0,0 +1,430 @@
...
+				Actions: []string{"$merge(\"merge\")"},
+			},
+		},
+	}
+
+	gotReviewpadFile, err := inlineImports(reviewpadFile, loadEnv)
+
+	assert.Nil(t, err)
+	assert.Equal(t, wantReviewpadFile, gotReviewpadFile)
+}
```
ferreiratiago, line 430: It's a bit confusing that this file is called `internal` and tests private and public functions. Also, may I suggest to convert this to not internal and use table testing?

```
func TestLoad() {
	"when parse error"
	"when success"
}

func TestLoad_Imports() {
	"when import file not found"
	"when import file is corrupt"
	"when import file fails to parse"
	"when import file is already loaded"
	"when import file imports other files"
	"when cyclic dependency"
}

func TestLoad_Transform() {
	"for actions"
	"for extra actions"
}
```

ferreiratiago: I would also suggest to create an input file to each one of the use cases on `testdata/loader`.

For instance, the following files could be added:
* reviewpad_with_parse_error.yml
* reviewpad_with_import_not_found.yml
* reviewpad_with_import_corrupted.yml
* ...
For the testing input maybe we can add a folder named `testdata/loader/transform` and add the files for input and outputs.
* reviewpad_transform_action.input.yml
* reviewpad_transform_action.output.yml

This will help to reduce the test sizes and to remove the added mocks on `mocks.go`.

WDYT?

ferreiratiago (changes requested):
My suggestion here is to refactor the tests taking into account table testing and the creating of file mocks. Please have a look at the comments.

engine/loader_internal_test.go
shay2025: @ferreiratiago firstly, thanks for the suggestions!
After thinking about it a bit:
1. Maybe we should apply the approach of `xyz_test.go` for external API and `xyz_internal_test.go` for internal API. WDYT?
2. Also I didn't quite understand the appoach you suggested for TDT. Shouldn't it be something like:
```
func TestParser() {
	"when error"
	"when success"
}

func TestTransform() {
      "for actions"
      "for extra actions"
}

func TestLoadImport() {
	"when import file not found"
	"when import file fails to parse"
	"when success"
}

func TestInlineImports() {
	"when import file not found"
	"when cyclic dependency"
        "when import file is already loaded"
        "when import file imports other files"
}

func TestLoad() {
	"when parse error"
	"when success"
}
```

ferreiratiago: My suggestion was not following `_internal_`. That's why all suggestion are `TestLoad_X`. 

I have thought about the tests answering the question **How should the loader behave based on the provided file?**, mainly:
* when the file has a parsing error
* when the file has an import of a non existing file
* when the file has an import of a corrupted file
* when the file has an import of a file with a parsing error
* when the file has chain of imports
* when the file has cyclic imports
* when the file has no issues

But this does not cover all cases. For instance, the use cases for transform do not get covered here. That's why I have suggested the following scenarios:
* when the file requires action transformation
* when the file requires extra action transformation

The next step was to organize these tests into categories, such as `_Imports` and `_Transform`.

For black box tests is not really important the `inlineImports` or the `transform` function. What's really important is that whenever I provided a file with a certain type of characteristics  (e.g. parsing error), the output will always be the same.

Saying that I think we should focus on how loader should behave regardless of its implementation details.

shay2025: Yes but we decided previously we would also test implementation that's why we were adding `_internal_` keyword into the test file name because by convention if you decide to do internal API tests you should introduce `_internal_` keyword to the test file name in order to flag them as being more brittle. What I would suggest indeed is to separe external API tests from internal API tests but this would create many more files.

engine/loader_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+			var wantReviewpadFile *engine.ReviewpadFile
+			isExpectedReviewpadFile := test.wantReviewpadFilePath != ""
+			if isExpectedReviewpadFile {
+				wantReviewpadFileData := httpmock.File(test.wantReviewpadFilePath).Bytes()
+				wantReviewpadFile = &engine.ReviewpadFile{}
+				yaml.Unmarshal(wantReviewpadFileData, &wantReviewpadFile)
+				// At the end of loading all imports from the file, its imports are reset to []engine.PadImport{}.
+				// However, the parsing of the wanted reviewpad file, sets the imports to []engine.PadImport(nil).
+				wantReviewpadFile.Imports = []engine.PadImport{}
```
shay2025, line 119: @ferreiratiago following the code comment hightlighted, shouldn't this:
https://github.com/reviewpad/reviewpad/blob/d867d9e1c1bedf9ac7041f266333d0beaf0d5fac/engine/loader.go#L165-L166
be:
```
	// reset all imports
	file.Imports = []PadImport(nil)
```

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+	type httpMockResponder struct {
+		url       string
+		responder httpmock.Responder
+	}
+
+	tests := map[string]struct {
+		httpMockResponders                                     []httpMockResponder
+		inputReviewpadFilePath, wantReviewpadFilePath, wantErr string
+	}{
+		// Fail
```
ferreiratiago, line 27: No need for this comment.

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+		"when the file has a parsing error": {
+			inputReviewpadFilePath: "../testdata/engine/loader/reviewpad_with_parse_error.yml",
+			wantErr:                "yaml: unmarshal errors:\n  line 5: cannot unmarshal !!str `parse-e...` into engine.ReviewpadFile",
+		},
+		"when the file has an import of a nonexisting file": {
```
ferreiratiago, line 32:
```suggestion
		"when the file imports a nonexistent file": {
```

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+			inputReviewpadFilePath: "../testdata/engine/loader/reviewpad_with_import_of_nonexisting_file.yml",
```
ferreiratiago, line 33:
```suggestion
			inputReviewpadFilePath: "../testdata/engine/loader/reviewpad_with_import_of_nonexistent_file.yml",
```

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+					responder: httpmock.NewBytesResponder(200, httpmock.File("../testdata/engine/loader/reviewpad_with_cyclic_dependency_b.yml").Bytes()),
+				},
+				{
+					url:       "https://foo.bar/reviewpad_with_cyclic_dependency_a.yml",
+					responder: httpmock.NewBytesResponder(200, httpmock.File("../testdata/engine/loader/reviewpad_with_cyclic_dependency_a.yml").Bytes()),
+				},
+			},
+			wantErr: "loader: cyclic dependency",
+		},
+		// Pass
```
ferreiratiago, line 66: No need for this comment.

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+		"when the file imports other files": {
```
ferreiratiago, line 67:
```suggestion
		"when the file has import chains": {
```

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+			wantReviewpadFilePath:  "../testdata/engine/loader/reviewpad_with_no_imports.yml",
+		},
+		"when the file requires action transformation": {
+			inputReviewpadFilePath: "../testdata/engine/loader/reviewpad_before_action_transform.yml",
+			wantReviewpadFilePath:  "../testdata/engine/loader/reviewpad_after_action_transform.yml",
+		},
+		"when the file requires extra action transformation": {
+			inputReviewpadFilePath: "../testdata/engine/loader/reviewpad_before_extra_action_transform.yml",
+			wantReviewpadFilePath:  "../testdata/engine/loader/reviewpad_after_extra_action_transform.yml",
+		},
```
ferreiratiago, line 92: These files should be inside a folder called `transform` under `testdata/engine/loader/transform`.

engine/loader_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			gotReviewpadFile, err := engine.Load(reviewpadFileData)
+
+			if err != nil && err.Error() != test.wantErr {
+				assert.FailNow(t, "Load() error = %v, wantErr %v", err, test.wantErr)
+			}
+			assert.Equal(t, wantReviewpadFile, gotReviewpadFile)
+		})
+	}
+}
```
ferreiratiago, line 130: This code is a bit confusing because we have to deal with both cases when there's error and don't. What do you think about splitting those cases into different loops?

```go
	for name, test := range tests {
		if test.wantErr == "" {
			continue
		}

		t.Run(name, func(t *testing.T) {
			// code for testing when error is expected
		})
	}

	for name, test := range tests {
		if test.wantErr != "" {
			continue
		}

		t.Run(name, func(t *testing.T) {
			// code for testing when no error is expected
		})
	}
```

testdata/engine/loader/reviewpad_after_action_transform.yml
```diff
@@ -0,0 +1,19 @@
...
+    description: Always true
+    spec: true
+
+workflows:
+  - name: test-process
+    description: Test process
+    if:
+      - rule: tautology
+    then:
+      - '$merge("merge")'
```
ferreiratiago, line 19: I understand the need of theses files and I'm all for it. My suggestion is to reduce the file content to only what it's actually needed.

```suggestion
api-version: reviewpad.com/v3.x

rules:
  - name: tautology
    kind: patch
    spec: true

workflows:
  - name: merge
    if:
      - rule: tautology
    then:
      - $merge("merge")
```

ferreiratiago: Please review all mock files to make sure we only have what's needed.

testdata/engine/loader/reviewpad_appended.yml
```diff
@@ -0,0 +1,53 @@
...
+    if:
+      - rule: is-medium
+    then:
+      - '$addLabel("medium")'
+  - name: auto-merge-owner-pull-requests
+    description: Auto merge pull requests
+    if:
+      - rule: auto-merge-authored-by-owners
+    then:
+      - '$merge("merge")'
```
ferreiratiago, line 53: Here's a good example. This file is too big. There's no need for that. Suggestion:

For the file `testdata/engine/loader/reviewpad_with_no_imports.yml`.

```
api-version: reviewpad.com/v3.x

rules:
  - name: is-medium
    kind: patch
    spec: $size() > 30 && $size() <= 100

workflows:
  - name: medium
    if:
      - rule: is-medium
    then:
      - $addLabel("medium")
```

And then for file `testdata/engine/loader/reviewpad_with_one_import.yml`

```yml
api-version: reviewpad.com/v3.x

imports:
  - url: https://foo.bar/reviewpad_with_no_imports.yml

rules:
  - name: is-large
    kind: patch
    spec: $size() > 100

workflows:
  - name: large
    if:
      - rule: is-large
    then:
      - $addLabel("large")
```

And then, this one become much simpler.

```yml
api-version: reviewpad.com/v3.x

rules:
  - name: is-medium
    kind: patch
    spec: $size() > 30 && $size() <= 100
  - name: is-large
    kind: patch
    spec: $size() > 100

workflows:
  - name: medium
    if:
      - rule: is-medium
    then:
      - $addLabel("medium")
  - name: large
    if:
      - rule: is-large
    then:
      - $addLabel("large")
```

We can't forget that each one of these files need to be maintained.

ferreiratiago (changes requested):
We are going into the right direction. Just need to improve the mock files to make sure we don't have unnecessary code.

engine/loader_test.go
shay2025: By that logic, we should also do the same for other internal functions that are being used in `Load`. I don't see the need for this.

engine/loader_test.go
shay2025: How is it confusing?
By that same logic, we should also do the same for the input and output reviewpad file when it is expected one and when it isn't.

testdata/engine/loader/reviewpad_appended.yml
shay2025: I made this file so big on purpose because here we are testing all the appends: `appendLabels`, `appendGroups` and `appendRules`. This is to take into account everything.

engine/loader_test.go
```diff
@@ -0,0 +1,162 @@
...
+			wantReviewpadFilePath:  "testdata/loader/transform/reviewpad_after_extra_action_transform.yml",
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			if test.httpMockResponders != nil {
+				// Starts the mock environment
+				httpmock.Activate()
+				// Shuts down the mock environment, removes the registered mocks and returns the mock environment to its initial state.
```
ferreiratiago, line 101: No need for these comments.

engine/testdata/loader/reviewpad_appended.yml
```diff
@@ -0,0 +1,51 @@
+# Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+# Use of this source code is governed by a license that can be
+# found in the LICENSE file.
+
+# We use this file for as mocked reviewpad file output for the test with the use case of when a file has import chains.
```
ferreiratiago, line 5: No need for this comment.

engine/testdata/loader/reviewpad_appended.yml
```diff
@@ -0,0 +1,51 @@
...
+# This file should result in the merge of the files:
```
ferreiratiago, line 6:
```suggestion
# This file is the result of merging the following files:
```

ferreiratiago (approved):
Some minor comments but all good 👍🏼

PR #180, opened 2022-07-27
test: add unit tests for engine exec

### LINKED ISSUE #176: Test Eval function in engine exec
To conclude a first phase of adding tests, we need to add unit tests for the Eval function.

shay2025:
## Description

This pull request introduces the unit tests for engine exec and some extra mocks related to the pull request data.
During the development of the unit tests, it was encountered an error in the `CollectError` function where it wasn't being taken into account the `*github.ErrorResponse` error type and this would cause an nil pointer dereference because the `Error()` method would be applied to this object which doesn't support such thing and so we need to manually get the error string in this case.

## Related issue

Closes #176

## Type of change

## How was this tested?

Ran `task test` cmd.

## Checklist

### DIFF
--- a/engine/mocks.go
+++ b/engine/mocks.go
@@ -66,15 +66,16 @@ func GetDefaultMockPullRequestDetails() *github.PullRequest {
 }
 
 func getDefaultMockPullRequestFileList() *[]*github.CommitFile {
-	prRepoName := defaultMockPrRepoName
 	return &[]*github.CommitFile{
 		{
-			Filename: github.String(fmt.Sprintf("%v/file1.ts", prRepoName)),
-			Patch: github.String(`@@ -2,9 +2,11 @@ package main
+			Filename: github.String(fmt.Sprintf("%v/file1.ts", defaultMockPrRepoName)),
+			Patch: github.String(
+				`@@ -2,9 +2,11 @@ package main
 - func previous1() {
 + func new1() {
 +
-return`),
+return`,
+			),
 		},
 	}
 }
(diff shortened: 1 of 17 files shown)


ferreiratiago (changes requested):
I would suggest to refactor the test and have into consideration table testing. Here's a starting point:
```
func TestEval_Labels() {
	"when label has no name"
	"when label exists"
	"when label does not exist"
	"when label is invalid"
}

func TestEval_Groups() {
	"when group is valid"
	"when group is invalid"
}

func TestEval_Rules() {
	"when rule is valid"
	"when rule is invalid"
}

func TestEval_Workflows() {
	"when workflow is invalid" // e.g. rule is invalid
	"when no workflow is activated"
	"when one workflow is activated"
	"when multiple workflows are activated"
	"when workflow should always run"
	"when workflow is activated and has extra actions"
}
```

You can check [hasUnaddressedReviewThreads_test.go](https://github.com/reviewpad/reviewpad/blob/main/plugins/aladino/functions/hasUnaddressedReviewThreads_test.go) to have an idea.

Also, I would suggest to have a different `reviewpad.yml` file for each test case in `testdata/exec`. For instance:
* reviewpad_with_unmaned_label.yml
* reviewpad_with_valid_label.yml (this can be used for `"when label does not exist"` and `"when label exists"`)
* ...
* reviewpad_with_one_worfklow
* reviewpad_with_multiple_worfklows

WDYT?

engine/mocks.go
```diff
@@ -35,17 +41,82 @@ func GetDefaultMockPullRequestDetails() *github.PullRequest {
...
+	}
+
+	mocks := append(clientOptions, defaultMocks...)
+
+	return github.NewClient(mock.NewMockedHTTPClient(mocks...))
 }
 
-func MockGithubClient(clientOptions []mock.MockBackendOption) *github.Client {
-	return github.NewClient(mockDefaultHttpClient(clientOptions))
+func MockDefaultEnvWith(client *github.Client, mockedAladinoInterpreter Interpreter) (*Env, error) {
```
ferreiratiago, line 103:
```suggestion
func MockEnvWith(client *github.Client, interpreter Interpreter) (*Env, error) {
```

ferreiratiago (changes requested):

ferreiratiago (approved):

PR #181, opened 2022-07-27
fix(cli): panic without optional flag `event-payload` string

### LINKED ISSUE #178: CLI panic without optional flag 'event-payload string'
**Describe the bug**
new CLI flag described as optional
```
  -event-payload string (optional)
        File path to github action event in JSON format
```

but if i run CLI without this flag i got panic in file
cmd/cli/main.go L_52

because here ev.Name & ev.Payload are nil

```
func parseEvent(rawEvent string) (interface{}, error) {
	ev := &Event{}

	err := json.Unmarshal([]byte(rawEvent), ev)
	if err != nil {
		return nil, err
	}

	return github.ParseWebHook(*ev.Name, *ev.Payload)
}
```

**To Reproduce**
Run CLI without this optional flag

**Expected behavior**
No errors

**Screenshots**
CleanShot 2022-07-27 at 16 54 45@2x

CleanShot 2022-07-27 at 16 54 45@2x

thedevsaddam:
## Description

Prevent panic while not providing optional `event-payload` string

## Related issue

Closes #178

## Type of change

## How was this tested?

I ran `task test -f`, `task build-cmd -f` commands

## Checklist


ferreiratiago:
Hi @thedevsaddam Thanks for your contribution 🙌🏼 The idea is to be able to run `reviewpad` cli without the event, i.e. an empty event. Using a default event will require to (1) support the default event file and (2) not guarantee that the cli fails whenever an event is really necessary. Therefore we will close this pull request in favor of #189

PR #192, opened 2022-07-28
feat: remove mocks files from code coverage

### LINKED ISSUE #191: Remove mocks files from code coverage
The mock files are currently being counted by go coverage tool and thus reduces the code coverage percentage for the packages where they are consumed.

Currently, we ignore the generated `parser.go` file from the code coverage by specifying the file path in `exclude-from-code-coverage.txt` which is consumed by a script (`exclude-from-code-coverage.sh`) that excludes this file from the code coverage.
We should add the file paths of the mocks files to this `exclude-from-code-coverage.txt`.

*None*

**Additional context**
*None*

shay2025:
## Description

This pull request introduces the removal of mock files in the code coverage by adding the file path pattern that matches `/mocks.go` to `exclude-from-code-coverage.txt` which is a file containing the file patterns regex which will be consumed by the `exclude-from-code-coverage.sh` script that removes from the code coverage the files that match the file patterns mentioned before.

## Related issue

Closes #191

## Type of change

## How was this tested?

Ran the `go tool cover -func coverage.out` that lists the files part of the code coverage with the covered percentage and it also presents at the end the total coverage percentage - with this it was verified that the parser and mock files didn't belong to this listing and the total coverage increased.

## Checklist

### DIFF
--- a/scripts/resources/exclude-from-code-coverage.txt
+++ b/scripts/resources/exclude-from-code-coverage.txt
@@ -1,1 +1,1 @@
-\/parser\.go\|\/mocks\.go
+\/\(parser\|mocks\)\.go
--- a/scripts/resources/exclude-from-code-coverage.txt
+++ b/scripts/resources/exclude-from-code-coverage.txt
@@ -1,1 +1,1 @@
-lang/aladino/parser.go
+\/parser\.go\|\/mocks\.go


scripts/resources/exclude-from-code-coverage.txt
```diff
@@ -1 +1 @@
-lang/aladino/parser.go
\ No newline at end of file
+\/parser\.go\|\/mocks\.go
```
marcelosousa, line 1: Please check if this simplification make sense:

```suggestion
\/[parser|mocks]\.go
```

shay2025: @marcelosousa this wont't work because square brackets are meant for character class, and you're actually trying to match any one of: `p`, `a`, `r`, `s`, `e`, `r` (again) | `m`, `o`, `c`, `k` and `s` (again).
What we can do to improve the regex is `\/\(parser\|mocks\)\.go`

marcelosousa (approved):

Issue #156, opened 2022-07-22
Use assert.FailNow in unit tests


marcelosousa:
**Is your feature request related to a problem? Please describe.**
At the moment, we are using log.Fatal to report failures in the unit tests.

**Describe the solution you'd like**
Use assert.FailNow.

marcelosousa:
Closed by #175

PR #175, opened 2022-07-27
change test method to assert.FailNow.

denizzengin:
## Related issue

#156 

## Type of change

Improvements (non-breaking change without functionality) 

## Checklist

- [x] I have performed a self-review of my code
- [x] I have run `task check -f` and have no issues

PR #194, opened 2022-07-29
refactor: fail on mockEnvWith function instead of returning error

ferreiratiago:
## Description

The function mocks function for `env` are being for all the tests and may return an error, which pollutes all tests with the famous `if err != nil`.

However, this can be avoided if the function itself fails, which is enough for the unit test to fail. This is what this pull request proposes.

The big change is related with the fact that mocked `env` touches all unit tests.

## Related issue

Related to #156 

## Type of change

Refactor (non-breaking change without functionality)

## How was this tested?

The command `task check -f` was run and the coverage was maintained.

## Checklist

### DIFF
--- a/lang/aladino/env_test.go
+++ b/lang/aladino/env_test.go
@@ -18,10 +18,7 @@ import (
 )
 
 func TestGetCtx_WithDefaultEnv(t *testing.T) {
-	mockedEnv, err := aladino.MockDefaultEnv(nil, nil)
-	if err != nil {
-		log.Fatalf("mockDefaultEnv failed: %v", err)
-	}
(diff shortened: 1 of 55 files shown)


marcelosousa:
@ferreiratiago I think for this refactor to make more sense, we should pass `*testing.T` to mockEnvWith. This is because the way we are failing at the moment is using `log.Fatal` which I don't think is the best. 

Please check https://github.com/reviewpad/reviewpad/issues/156 and https://github.com/reviewpad/reviewpad/pull/175. 

WDYT?

ferreiratiago:
@marcelosousa You're right. Also, totally miss the issue and the pull request 🤦🏼‍♂️ I'll update this pull request and sync with @denizzengin

PR #197, opened 2022-07-29
feat/builtin unresolvedThreadCount

ferreiratiago:
- refactor: fail on mockEnvWith function instead of returning error
- feat: Add builtin unResolvedThreadCount

## Checklist


plugins/aladino/functions/unresolvedThreadCount_test.go
```diff
@@ -0,0 +1,122 @@
...
+									Repo: &github.Repository{
+										Owner: &github.User{
+											Login: github.String(mockedPrOwner),
+										},
+										Name: github.String(mockedPrRepoName),
+									},
+								},
+							}),
+						))
+					}),
```
shay2025, line 88: I don't think it is needed to overwrite this request.
After some thought, I believe it only makes sense to overwrite it when we want to change the values of the pull request parameters which isn't the case because they are already defined in the mocks files.
I thought maybe you were overwriting it with the thought of a possible scenario that this parameters didn't exist in the current mocked pull request.
However, I believe the correct procedure would be:
1. Check in the mock files if the relevant parameters to the test are set in the mocks - if not, mock them.
2. If the test is testing the values of the parameters, overwrite the get pull request request.

plugins/aladino/functions/unresolvedThreadCount_test.go
```diff
@@ -0,0 +1,122 @@
...
+				}
+			},
+		)
+
+		t.Run(name, func(t *testing.T) {
+			gotVal, err := unresolvedThreadCount(mockedEnv, []aladino.Value{})
+
+			assert.Nil(t, err)
+			assert.Equal(t, test.wantVal, gotVal)
+
```
shay2025, line 119: Remove this empty line please :)

utils/github.go
```diff
@@ -301,3 +333,39 @@ func GetPullRequests(ctx context.Context, client *github.Client, owner string, r
...
+				continue
+			} else {
+				return nil, fmt.Errorf("GetReviewThreads fatal: %v", err)
+			}
+		} else {
+			currentRequestRetry = 0
+		}
+
+		reviewThreads = append(reviewThreads, reviewThreadsQuery.Repository.PullRequest.ReviewThreads.Nodes...)
+		hasNextPage = reviewThreadsQuery.Repository.PullRequest.ReviewThreads.PageInfo.HasNextPage
```
shay2025, line 366: Instead of reassigning this var maybe we should perform this check inside the loop and break if needed since this would avoid unnecessary code line executions like the next one (L367).

shay2025 (approved):
Minor thoughts

utils/github.go
```diff
@@ -6,14 +6,46 @@ package utils
...
+	Comments struct {
+		TotalCount githubv4.Int
+		Nodes      []struct {
+			Author GQLActor
+		}
+	} `graphql:"comments(first: 1)"`
+	IsOutdated githubv4.Boolean
+	IsResolved githubv4.Boolean
+	Path       githubv4.String
+}
```
marcelosousa, line 33: We shouldn't retrieve more information than required:

```suggestion
type GQLReviewThread struct {
	IsResolved githubv4.Boolean
}
```

utils/github.go
```diff
@@ -301,3 +333,39 @@ func GetPullRequests(ctx context.Context, client *github.Client, owner string, r
...
+		"pullRequestNumber":   githubv4.Int(number),
+		"reviewThreadsCursor": (*githubv4.String)(nil),
+	}
+
+	currentRequestRetry := 1
+
+	for hasNextPage {
+		err := client.Query(context.Background(), &reviewThreadsQuery, varGQLReviewThreads)
+		if err != nil {
+			fmt.Printf("try %v, graphQL GetReviewThreads %v\n", currentRequestRetry, err)
```
marcelosousa, line 354: For consistency, I think we should log instead of fmt.

marcelosousa (changes requested):

plugins/aladino/functions/hasUnaddressedThreads.go
```diff
@@ -0,0 +1,36 @@
...
+		Code: hasUnaddressedThreadsCode,
+	}
+}
+
+func hasUnaddressedThreadsCode(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
+	prNum := utils.GetPullRequestNumber(e.GetPullRequest())
+	owner := utils.GetPullRequestBaseOwnerName(e.GetPullRequest())
+	repo := utils.GetPullRequestBaseRepoName(e.GetPullRequest())
+
+	reviewThreads, err := utils.GetReviewThreads(e.GetCtx(), e.GetClientGQL(), owner, repo, prNum, 2)
```
marcelosousa, line 24: It's not clear that `2` is the number of tries. Please add a variable to make it explicit.

utils/github.go
```diff
@@ -301,3 +323,39 @@ func GetPullRequests(ctx context.Context, client *github.Client, owner string, r
 
 	return prs.([]*github.PullRequest), nil
 }
+
+func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string, repo string, number int, retryCount int) ([]GQLReviewThread, error) {
+	var reviewThreads []GQLReviewThread
```
marcelosousa, line 328: I think it's best to do:
```
reviewThreads := make([]GQLReviewThread, 0)
```

The reason is that with `var reviewThreads []GQLReviewThread`, `reviewThreads` is nil. If for some reason, we change the logic so that nothing is added to the slice if there are no threads, we'll ending up returning `nil, nil` which might lead to a nil pointer dereference in the callers.

marcelosousa (changes requested):

marcelosousa (approved):
Leaving the issue of the long query strings for a later issue.

PR #221, opened 2022-08-03
fix: fetch last update date from push for isWaitingForReview

### LINKED ISSUE #208: Fetch last update date from push for $isWaitingForReview
Right now the builtin `$isWaitingForReview` requires the value of the **last update date** in order to understand if a pull request is waiting for review or not.

This value is being fetched from list of the pull request commits, mainly from the last commit. This is because the list of commits is in chronological order.

However, it is possible for a user to set a commit date manually, which means that we can't trust the commit date.

The proposed solution is to look for the date of the last pull request push.

_none_

**Additional context**
Fetching the last push it means to fetch all push request timeline events and filter all events in order to get the last push.

The timeline fetch is only possible through github graphql api.

pkbhowmick:
Signed-off-by: Pulak Kanti Bhowmick <pkbhowmick007@gmail.com>

## Related issue
https://github.com/reviewpad/reviewpad/issues/208

## Checklist


utils/github.go
```diff
@@ -34,6 +36,21 @@ type ReviewThreadsQuery struct {
...
+				Nodes []struct {
+					Commit struct {
+						Oid        string
+						PushedDate string
+					}
+				}
+			} `graphql:"commits(last: 1)"`
+		} `graphql:"pullRequest(number: $pullRequestNumber)"`
+	} `graphql:"repository(owner: $repositoryOwner, name: $repositoryName)"`
+}
```
ferreiratiago, line 52: This query does not cover all cases. It solves the problem of someone changing the commit date but it misses any force pushes. My suggestion is to query github timeline for `HEAD_REF_FORCE_PUSHED_EVENT` and `PULL_REQUEST_COMMIT`.

Here's the query for [graphiql](https://graphql-dotnet.github.io/docs/getting-started/graphiql/).
```graphql
{
  repository(owner: "ferreiratiago", name: "paddy") {
    pullRequest(number: 4) {
      timelineItems(last: 2, itemTypes: [HEAD_REF_FORCE_PUSHED_EVENT, PULL_REQUEST_COMMIT]) {
        nodes {
          ... on HeadRefForcePushedEvent {
            createdAt
          }
          ... on PullRequestCommit {
            commit {
              pushedDate
            }
          }
        }
      }
    }
  }
}
```

_(The query is prepared for a personal example project)_

WDYT?

utils/github.go
```diff
@@ -34,6 +36,21 @@ type ReviewThreadsQuery struct {
 	} `graphql:"repository(owner: $repositoryOwner, name: $repositoryName)"`
 }
 
+type PRLastPushedDateQuery struct {
```
ferreiratiago, line 39: Suggestion name `LastPushedEventDateQuery`

utils/github.go
```diff
@@ -356,3 +373,22 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
 
 	return reviewThreads, nil
 }
+
+func GetPRLastPushedCommitDate(ctx context.Context, client *githubv4.Client, owner string, repo string, number int) (time.Time, error) {
```
ferreiratiago, line 377: In order to maintain consistency let's call this `GetPullRequestLastPushedEventDate`.

utils/github.go
```diff
@@ -356,3 +373,22 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
...
+	var pullRequestQuery PRLastPushedDateQuery
```
ferreiratiago, line 378: Naming suggestion `lastPushedEventDateQuery`.

plugins/aladino/functions/isWaitingForReview.go
```diff
@@ -43,13 +43,16 @@ func isWaitingForReviewCode(e aladino.Env, _ []aladino.Value) (aladino.Value, er
 		return aladino.BuildBoolValue(false), nil
 	}
 
-	// FIXME: #208
 	lastCommit := commits[len(commits)-1]
 	if lastCommit.Commit == nil || lastCommit.Commit.Committer == nil || lastCommit.Commit.Committer.Date == nil {
 		log.Printf("[WARN] Commit %v has no value for pull request %s/%s#%d.", lastCommit, owner, repo, prNum)
 		return aladino.BuildBoolValue(false), nil
 	}
```
ferreiratiago, line 50: I'm afraid this is not required anymore.

ferreiratiago (changes requested):
Thanks for the contribution @pkbhowmick 🙌🏼 It's looking great. We need to update the graphql query to also cover the force push case. Please have a look at the comments. Thanks

utils/github.go
pkbhowmick: yes, I also think so.

utils/github.go
```diff
@@ -34,6 +37,25 @@ type ReviewThreadsQuery struct {
...
+					HeadRefForcePushedEvent struct {
+						CreatedAt string
+					} `graphql:"... on HeadRefForcePushedEvent"`
+					PullRequestCommit struct {
+						Commit struct {
+							PushedDate string
+						}
+					} `graphql:"... on PullRequestCommit"`
+				}
+			} `graphql:"timelineItems(last: 2, itemTypes: [HEAD_REF_FORCE_PUSHED_EVENT, PULL_REQUEST_COMMIT])"`
```
ferreiratiago, line 54: No need for the last 2. My bad that suggested that. One is enough - the last one :)

utils/github.go
```diff
@@ -34,6 +37,25 @@ type ReviewThreadsQuery struct {
 	} `graphql:"repository(owner: $repositoryOwner, name: $repositoryName)"`
 }
 
+type LastPushedEventDateQuery struct {
```
ferreiratiago, line 40: Now that we iterate again I was thinking if we shouldn't call this struct `LastPushQuery`. If we think about it `pushed` doesn't add more value, `event` is just a detail and `date` is one of the fields. We may even add more fields and having `date` in the name will mess that up. WDYT?

utils/github.go
```diff
@@ -356,3 +378,46 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
 
 	return reviewThreads, nil
 }
+
+func GetPullRequestLastPushedEventDate(ctx context.Context, client *githubv4.Client, owner string, repo string, number int) (time.Time, error) {
```
ferreiratiago, line 382: I think we could rename this to `GetPullRequestLastPushDate`. Saying `PushedEvent` does not really helps to understand what the function does.

utils/github.go
```diff
@@ -356,3 +378,46 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
...
+	var lastPushedEventDateQuery LastPushedEventDateQuery
+	varGQLastPushedEventDate := map[string]interface{}{
```
ferreiratiago, line 384: These get updated to `lastPushQuery` and `varGQLastPush`.

utils/github.go
```diff
@@ -356,3 +378,46 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
...
+		"pullRequestNumber": githubv4.Int(number),
+	}
+
+	err := client.Query(ctx, &lastPushedEventDateQuery, varGQLastPushedEventDate)
+	if err != nil {
+		return time.Time{}, err
+	}
+	if len(lastPushedEventDateQuery.Repository.PullRequest.TimelineItems.Nodes) == 0 {
+		return time.Time{}, errors.New("pushedDate not found")
+	}
```
ferreiratiago, line 396: Let's make more verbose here and also add some lines between.

```suggestion
	
	hasLastPush := len(lastPushQuery.Repository.PullRequest.TimelineItems.Nodes) > 0
	if  !hasLastPush {
		return time.Time{}, errors.New("last push not found")
	}
	
```

utils/github.go
```diff
@@ -356,3 +378,46 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
...
+			pushedEventDates = append(pushedEventDates, pushedDate)
+		}
+	}
+	if len(pushedEventDates) == 0 {
+		return time.Time{}, errors.New("pushedDate not found")
+	}
+	sort.Slice(pushedEventDates, func(i, j int) bool {
+		return pushedEventDates[i].After(pushedEventDates[j])
+	})
+	return pushedEventDates[0], nil
```
ferreiratiago, line 422: I'm not sure if we need all this. Since we are fetching only one event, or even fetching multiple, we could extract the last one.

However, we need to make sure that the order is always the correct one, i.e. that the last event was in fact the last event to happen. Can you please check if that's the case? i.e. If github graphql API guarantees that the order is the correct one?

If that's the case I believe we can simplify this logic :)

ferreiratiago (changes requested):
We are going into the right direction. Please have a look at the comments.

utils/github.go
pkbhowmick: Can you please check if that's the case?
yes, I checked it. The github api returns the actual last event.

utils/github.go
```diff
@@ -356,3 +377,34 @@ func GetReviewThreads(ctx context.Context, client *githubv4.Client, owner string
...
+	if !hasLastPush {
+		return time.Time{}, errors.New("last push not found")
+	}
+	var pushDate string
+	node := lastPushQuery.Repository.PullRequest.TimelineItems.Nodes[0]
+	if node.PullRequestCommit.Commit.PushedDate != "" {
+		pushDate = node.PullRequestCommit.Commit.PushedDate
+	} else if node.HeadRefForcePushedEvent.CreatedAt != "" {
+		pushDate = node.HeadRefForcePushedEvent.CreatedAt
+	}
```
ferreiratiago, line 404: We can use graphql `__typename` for this. Can you please have a look?

ferreiratiago (changes requested):
Please use graphql `__typename` to choose which event to handle. Also, please remove the pull request from draft and update the title to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

marcelosousa:
@ferreiratiago: bumping the priority in this pull request.

ferreiratiago (approved):
This pull request is to be continue on the following issues:
- #331 
- #332

Issue #162, opened 2022-07-25
Normalize Reviewpad.yml


ferreiratiago:
**Is your feature request related to a problem? Please describe.**
Right now reviewpad configuration is done through `reviewpad.yml` file.

This configuration has optional properties. For instance, the property `mode`. This means that the value for the property `mode` can be filled (e.g. `mode: silent`) or not included in the configuration.

However, when no configuration is provided, reviewpad needs to assume a default behavior. In the case of `mode` the default behavior is `silent`.

We need to make sure that whenever an optional configuration is not provided, a default configuration is assumed.

**Describe the solution you'd like**
The idea is to create a normalize function able to normalize the `reviewpad.yml` configuration and make sure all optional parameters have a default value, i.e. `normalize(reviewpad.yml): normalized_reviewpad.yml`.

**Describe alternatives you've considered**
An alternative will be to have some read function that is able to safely read a property from `reviewpad.yml`. This function will require to either
- (1) know the default values for each optional property, i.e. `safely_read_property(reviewpad.yml, property): property_value` or 
- (2) ask for a default value in case the property does not exists, i.e. `safely_read_property(reviewpad.yml, property, default_value): property_value`.

**Additional context**
We need to make sure where is the best place to perform this normalization. Either on [reviewpad](https://github.com/reviewpad/reviewpad), [action](https://github.com/reviewpad/action) or [host-event-handler](https://github.com/reviewpad/host-event-handler).

PR #223, opened 2022-08-04
Normalize property values

sheshenia:
## Description

This is the solution N1 for the normalization property values issue.
The advantage of this solution is that it doesn't affect the existing codebase and is based on abstractions and interfaces.
The idea is to inject normalization functionality on YAML unmarshalling stage ./engine/unmarshallers.go
The "Normalization" functionality itself is composable for different modifications and validations and concentrated in one place. 

My solution N2 is partially based on N1 but will provide more flexibility and control on "getting property values" stage. But it will affect on existing codebase. N2 will be soon.

## Related issue

#162

## Type of change

## How was this tested?

Unit tests, TestNormalization in the ./engine/normalize_test.go

## Checklist


marcelosousa:
@sheshenia many thanks for this contribution 🙌  Please address the reviewpad-bot comments so that we can move on to the review phase 👍

Issue #195, opened 2022-07-29
Support string interpolation


shay2025:
**Is your feature request related to a problem? Please describe.**
Currently, we don't support string interpolation in Aladino and we may want to be able to do something like `$infof("Welcome %v", $author())`.

**Describe the solution you'd like**
I propose we create a new built-in called sprintf which receives two argumens:
- `format`: this argument is of type string and represents the string containing custom specifiers that the sprintf uses to format the final output string.
- `a ...interface{}`: this argument refers to the list of a variable number of arguments that need to be formatted and printed. 
And returns the formatted string.

So in Aladino terms, the built-in would have the type: `aladino.FunctionType([]aladino.Type{aladino.StringType()}, aladino.BuildArrayOfType(aladino.BuildStringType()))`.

Example of usage:
`$info(sprintf("Welcome %v", [$author()]))`

**Describe alternatives you've considered**
*None*

**Additional context**
*None*

PR #225, opened 2022-08-04
feat: sprintf returns formated string

versoul:
## Description

Feat adds formated string buit-in

## Related issue

Closes [issue #195](https://github.com/reviewpad/reviewpad/issues/195)

## Type of change

## How was this tested?
Ran `task test` command.

## Checklist

### DIFF
--- a/plugins/aladino/functions/sprintf.go
+++ b/plugins/aladino/functions/sprintf.go
@@ -1,0 +1,32 @@
+// Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+// Use of this source code is governed by a license that can be
+// found in the LICENSE file.
+
+package plugins_aladino_functions
+
+import (
+	"fmt"
+	"github.com/reviewpad/reviewpad/v3/lang/aladino"
+)
+
+func Sprintf() *aladino.BuiltInFunction {
+	return &aladino.BuiltInFunction{
+		Type: aladino.BuildFunctionType([]aladino.Type{aladino.BuildStringType(), aladino.BuildArrayOfType(aladino.BuildStringType())}, aladino.BuildStringType()),
+		Code: sprintfCode,
+	}
+}
+
+func sprintfCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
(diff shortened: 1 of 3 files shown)


plugins/aladino/functions/sprintf.go
```diff
@@ -0,0 +1,38 @@
...
+		case *aladino.StringValue:
+			clearVals = append(clearVals, val.(*aladino.StringValue).Val)
+		case *aladino.IntValue:
+			clearVals = append(clearVals, val.(*aladino.IntValue).Val)
+		case *aladino.BoolValue:
+			clearVals = append(clearVals, val.(*aladino.BoolValue).Val)
+		case *aladino.TimeValue:
+			clearVals = append(clearVals, val.(*aladino.TimeValue).Val)
+		}
+	}
```
marcelosousa, line 35: In L14, it is specified that the type of `sprintf` is a string plus an array of strings. This means that it's not possible for a `val` to be a IntValue as the type checker will complain. For now, it's best to simplify and assume that we can only pass vals of type String.

plugins/aladino/functions/sprintf_test.go
```diff
@@ -0,0 +1,39 @@
...
+			aladino.BuildStringValue("world"),
+		},
+	)
+	args := []aladino.Value{format, slice}
+
+	wantString := &aladino.StringValue{Val: "Hello, 1 big world!"}
+
+	gotString, err := sprintf(mockedEnv, args)
+	fmt.Println("GOT")
+	fmt.Println(gotString)
```
marcelosousa, line 35: This seem like debug code.

marcelosousa (changes requested):

plugins/aladino/functions/sprintf.go
versoul: it was for a future, but fixed

plugins/aladino/functions/sprintf_test.go
versoul: fixed

marcelosousa (approved):

PR #233, opened 2022-08-05
experimental: add pipelines

### LINKED ISSUE #231: Introduce pipelines

For certain use cases, the initial of workflows as a basic `if X then Y` is not appropriated. 

One example of this is when we want to enforce a pull request to be reviewed first by person X and only when person X approves, then a review is requested to person Y.

At the moment, to model such case, we could do it with the following workflow:

```yml
rules:
  - name: initial-condition
    spec: '....'

  - name: person-X-approved
    spec: '$reviewerStatus("X") == "APPROVED"'

workflows:
  - name: sequential-reviewer
    description: Ask for sequential reviewers
    if:
      - rule: initial-condition
        extra-actions:
          - '$assignReviewer(["X"])'
      - rule: person-X-approved
        extra-actions:
          - '$assignReviewer(["Y"])'  
```

Lets consider when a pull request where the initial-condition is true so that we can go through this review workflow. As soon as `$assignReviewer(["X"])` is executed, we want to be in a state where we are waiting for "X" approval.

When that happens, both the initial-condition and person-X-approved will be true. In this case, the executed program will:

```yml
$assignReviewer(["X"])
$assignReviewer(["Y"])
```

whereas our intention is to just execute `$assignReviewer(["Y"])`.

At first, I thought we could add a new built-in: `$sequentialReviewers([["X"], ["Y]], "APPROVED")` which receives a sequence of reviewer lists and advances to the next element of the sequence if everyone in the current list has "APPROVED". However, this solution seems very brittle, i.e. what if the user wants to specify a more complicated review process?

Therefore, I propose an extension of the current YML specification to support the concept of 'pipeline'.
```yaml
pipelines:
  - name: STRING
    trigger: [STRING | RULE] # optional
    stages:
      - actions:
           - ALADINO_ACTION_1
           - ALADINO_ACTION_2
        until: [STRING | RULE] # optional
       ...
      - actions:
           - ALADINO_ACTION_1
           - ALADINO_ACTION_2
        until: [STRING | RULE] # optional

```

A pipeline is running iff the trigger is the empty or its evaluation is true.

When a pipeline is running, only one of the stages will be executed.

This stage will be the first where its until clause is empty or its evaluation is true.

Going back to our use case of the sequence of reviewers, we could now specify it with a pipeline:

```yaml
pipelines:
  - name: sequential-reviewer
    trigger: initial-condition
    stages:
      - actions:
           - '$assignReviewer(["X"])'
        until: person-X-approved
      - actions:
           - '$assignReviewer(["Y"])'
```

marcelosousa:
## Description

This pull request introduces the concept of pipeline as an extension of our current specification.

## Related issue

Closes #231.

## Type of change

## How was this tested?

This feature was manually tested with first basic experiments using the CLI.

## Checklist

### DIFF
--- a/engine/exec.go
+++ b/engine/exec.go
@@ -160,14 +160,19 @@ func Eval(file *ReviewpadFile, env *Env) (*Program, error) {
 	for _, pipeline := range file.Pipelines {
 		execLogf("evaluating pipeline %v:", pipeline.Name)
 
-		activated, err := interpreter.EvalExpr("patch", pipeline.Trigger)
-		if err != nil {
-			CollectError(env, err)
-			return nil, err
+		var err error
+		activated := pipeline.Trigger == ""
+		if !activated {
+			activated, err = interpreter.EvalExpr("patch", pipeline.Trigger)
+			if err != nil {
+				CollectError(env, err)
+				return nil, err
+			}
 		}
 
 		if activated {
-			for _, stage := range pipeline.Stages {
+			for num, stage := range pipeline.Stages {
+				execLogf("evaluating pipeline stage %v", num)
 				if stage.Until == "" {
 					program.append(stage.Actions, pipeline.Name)
 					break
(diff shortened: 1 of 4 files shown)


ferreiratiago (approved):
Let's go 💪🏼

PR #259, opened 2022-08-11
feat: built-ins services

marcelosousa:
## Description

This pull request extends the plugins datastructure to be able to allow for dedicated services that can be used from the built-in services and actions.

Additionally, it starts the process of adding a first service for the reviewpad semantic service that will be used by the $hasAnnotation built-in.

## Type of change

## How was this tested?

Ongoing

## Checklist


plugins/aladino/actions/addLabel_test.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var addLabel = plugins_aladino.PluginBuiltIns().Actions["addLabel"].Code
+var addLabel = plugins_aladino.PluginBuiltIns(plugins_aladino.DefaultPluginConfig()).Actions["addLabel"].Code
```
marcelosousa, line 20: All the changes in the tests are similar to this one because we need to pass a configuration to create the PluginBuiltIns. 

**This might a bit problematic because this means that we will be creating a new semantic client for each test! **

plugins/aladino/services/semantic.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/reviewpad/api/go/clients"
+	"github.com/reviewpad/api/go/services"
+)
+
+func NewSemanticService(endpoint string) services.SemanticClient {
+	client, _ := clients.NewSemanticClient(endpoint)
+
+	return client
+}
```
marcelosousa, line 16: This is just a wrapper. Not happy with this solution because the client could be nil. I think the best way to improve this would be to directly inline this code in  `func PluginBuiltIns(config *PluginConfig) *aladino.BuiltIns` and perhaps remove this file altogether.

marcelosousa (commented):
Just added a few comments in case @shay2025 or @ferreiratiago pick this up while I'm on holidays.

Issue #268, opened 2022-08-19
Builtin $size should receive a list of glob patterns to be ignored


ferreiratiago:
**Is your feature request related to a problem? Please describe.**
Currently, the builtin `size` does not take any parameter. This can be a problem when a pull request adds, for instance, generated files.

Therefore, the builtin should be able to receive a list of patterns to be ignored when calculating the `size`.

**Describe the solution you'd like**
Update the builtin to receive a list of strings where each string represents a [glob pattern](https://en.wikipedia.org/wiki/Glob_(programming)) that should be ignored.

**Example:**
For the list of changed files:
- a.json
- b.go
- myGenFile.go

And builtin:
```yml
$size(["*.json", "*gen*"])
```

It should only count the changed files of:
- b.go

Issue #89, opened 2022-07-08
Support multiline values


ferreiratiago:
**Is your feature request related to a problem? Please describe.**
Right now our parser is not able to build the AST of a [multiline values](https://yaml-multiline.info/).

Here's an example:
```yml
workflows:
  - name: my-policy
    if:
      - rule: tautology
    then:
      - > 
        $warn("Required rules: 
        - rule 1
        - rule 2")
```

**Describe the solution you'd like**
Update parser in order to be able to handle multiline values.

**Describe alternatives you've considered**
_none_

**Additional context**
With multiline parse fails with error.

```sh
syntax error on 
2022/07/07 17:13:49 parse error: failed to build AST on input  $warn("Required rules:  - rule 1  - rule 2")
```

PR #269, opened 2022-08-20
feat: add patterns to size built-in

versoul:
## Description

Adds patterns to Size built-in

## Related issue

Closes [issue #89](https://github.com/reviewpad/reviewpad/issues/268)

## Type of change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

!!! now all Size built-in expects empty arrays as pattern, it wouldn't work without params as earlier

## How was this tested?

run tests and try to run reviewpad as a CLI with new config

## Checklist

### DIFF
--- a/lang/aladino/parser.go
+++ b/lang/aladino/parser.go
@@ -74,7 +74,7 @@ const AladinoInitialStackSize = 16
 
 /*  start  of  programs  */
 
-var AladinoExca = [...]int{
+var AladinoExca = [...]int8{
 	-1, 1,
 	1, -1,
 	-2, 0,
@@ -107,13 +107,13 @@ var AladinoPgo = [...]int{
 	0, 0, 2, 23,
 }
 
-var AladinoR1 = [...]int{
+var AladinoR1 = [...]int8{
 	0, 3, 1, 1, 1, 1, 1, 1, 1, 1,
 	1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
 	2, 2, 2,
 }
 
-var AladinoR2 = [...]int{
+var AladinoR2 = [...]int8{
 	0, 1, 2, 3, 3, 3, 3, 3, 3, 1,
 	1, 1, 1, 3, 2, 1, 1, 5, 5, 2,
 	3, 1, 0,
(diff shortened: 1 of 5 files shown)


marcelosousa (changes requested):
Thanks for this contribution @versoul! 🙌  

This solution is not backwards compatible - all users that were previously using `$size()` will now have to update to something like `$size(".*")`. 

We need to do something similar to what we did with the `$merge()` action - check the transform.go file.

versoul:
@marcelosousa fixed

lang/aladino/parser.go
```diff
@@ -69,7 +69,7 @@ const AladinoInitialStackSize = 16
 
 /*  start  of  programs  */
 
-var AladinoExca = [...]int{
```
marcelosousa, line 72: It is strange that this file was modified - there should be no changes in this file for this feature :(

plugins/aladino/functions/size.go
```diff
@@ -5,23 +5,58 @@
...
 	}
 }
 
-func sizeCode(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
-	pullRequest := e.GetTarget().(*target.PullRequestTarget).PullRequest
+func sizeCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
+	if len(args) == 0 {
+		size := e.GetTarget().(*target.PullRequestTarget).PullRequest.GetAdditions() + e.GetTarget().(*target.PullRequestTarget).PullRequest.GetDeletions()
+		return aladino.BuildIntValue(size), nil
+	}
```
marcelosousa, line 27: This conditional is dead code because in L17 it is specified that size always receives at least one argument.

marcelosousa (changes requested):
This feature is still not working as intended. Please check the comment. As I mentioned before, we'll need to change the file transform.go so that if $size does not have arguments, it will add the empty array as an argument.

lang/aladino/parser.go
versoul: i think this file changed every time when i run tests...

plugins/aladino/functions/size.go
versoul: this condition for case when we pass empty array as exclude param $size([]) and it works and needed, i rechecked again with this line and without

versoul:
@marcelosousa done

marcelosousa (approved):

PR #274, opened 2022-08-23
feat: inline rules with extra actions

### LINKED ISSUE #267: Allow to add extra actions on inline rules
Right now inline rules are allowed with the following syntax:

```yml
workflows:
  - name: size-small
    if:
      - "$size < 3"   # inline rule
    then:
      - $addLabel("small")
  
  - name: size-medium # inline rule
    if: 
      - "$size >= 3 && $size < 5"
    then:
      - $addLabel("medium")
  
  - name: size-large  # inline rule
    if:
      - "$size > 5"
    then:
      - $addLabel("large")
```

Although this solution works fine we want it to work with `extra-actions`.

We should be able to also define an inline rule with `rule` and add `extra-actions`.

Therefore the following workflow `size` should be allowed:
```yml
workflows:
  - name: size
    if:
      - rule: "$size < 3"               # inline rule
        extra-actions:
          - $addLabel("small")
      - rule: "$size >= 3 && $size < 5" # inline rule
        extra-actions:
          - $addLabel("medium")
      - rule: "$size > 5"               # inline rule
        extra-actions:
          - $addLabel("large")
```

zolamk:
## Description

This PR introduces support for inline rules with extra actions like below

```
workflows:
  - name: size
    if:
      - rule: "$size < 3"               # inline rule
        extra-actions:
          - $addLabel("small")
      - rule: "$size >= 3 && $size < 5" # inline rule
        extra-actions:
          - $addLabel("medium")
      - rule: "$size > 5"               # inline rule
        extra-actions:
          - $addLabel("large")
 ```
How this is implemented is during the loading stage we check if the rule exists in the Reviewpad file and if not we add it to the list of registered rules and refer the newly registered rule in the workflow

## Related issue

Closes #267

## Type of change

## How was this tested?
Unit Tests and Manual tests

## Checklist

### DIFF
--- a/engine/loader.go
+++ b/engine/loader.go
@@ -222,9 +222,10 @@ func processInlineRules(file *ReviewpadFile) (*ReviewpadFile, error) {
 		Labels:       file.Labels,
 		Workflows:    file.Workflows,
 	}
+	alreadyGeneratedRules := map[string]bool{}
 
 	for i, workflow := range reviewpadFile.Workflows {
-		processedWorkflow, rules, err := processInlineRulesOnWorkflow(workflow, file.Rules)
+		processedWorkflow, rules, err := processInlineRulesOnWorkflow(workflow, file.Rules, alreadyGeneratedRules)
 		if err != nil {
 			return nil, err
 		}
@@ -236,7 +237,7 @@ func processInlineRules(file *ReviewpadFile) (*ReviewpadFile, error) {
 	return reviewpadFile, nil
 }
 
-func processInlineRulesOnWorkflow(workflow PadWorkflow, currentRules []PadRule) (*PadWorkflow, []PadRule, error) {
+func processInlineRulesOnWorkflow(workflow PadWorkflow, currentRules []PadRule, alreadyGeneratedRules map[string]bool) (*PadWorkflow, []PadRule, error) {
 	wf := &PadWorkflow{
 		Name:        workflow.Name,
 		Description: workflow.Description,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: engine/loader.go [hunk 1]
```diff
@@ -242,24 +242,34 @@ func processInlineRulesOnWorkflow(workflow PadWorkflow) (*PadWorkflow, []PadRule
...
+				Rule: padRule.Name,
 			})
 		case map[string]interface{}:
 			rule, err := mapToPadWorkflowRule(r)
 			if err != nil {
 				return nil, nil, err
 			}
 
+			// if the rule doesn't exist in the list of rules
+			// treat it as an inline rule with extra actions
```
ferreiratiago, line 259:
```suggestion
			// if the rule doesn't exist in the list of available rules
			// treat it as an inline rule
```

engine/loader.go
```diff
@@ -242,24 +242,34 @@ func processInlineRulesOnWorkflow(workflow PadWorkflow) (*PadWorkflow, []PadRule
...
+			if _, exists := findRule(rules, rule.Rule); !exists {
+				padRule := inlineRuleToPadRule(rule.Rule)
+
+				foundInlineRules = append(foundInlineRules, padRule)
+
+				wf.Rules = append(wf.Rules, PadWorkflowRule{
+					Rule:         padRule.Name,
+					ExtraActions: rule.ExtraActions,
```
ferreiratiago, line 267: Having extra action is not mandatory. Please make sure that this works without `extra-actions`.

e.g.
```yml
workflows:
  - name: test-process
    if:
      - rule: '$size() <= 30'
        extra-actions:
          - $addLabel("small")
      - rule: '$size() > 30'
    then:
      - '$addLabel("critical")'
```

ferreiratiago (changes requested):
Didn't look into the tests but please go ahead with the suggested recommendations.

engine/loader.go
zolamk: Yeah, I've checked and confirmed it works properly

ferreiratiago (approved):

PR #278, opened 2022-08-23
fix: retrieve creation time when timeline has only one event

### LINKED ISSUE #277: `$lastEventAt` retrieves wrong time for PRs with only one event
**Describe the bug**
The `$lastEventAt` built-in in order to retrieve the last event time, performs a request to list the issue timeline.

However, when we only have only an event in this timeline (the creation of the PR) the returned time is wrong returning a negative number since the time is not provided for the first element for unknown reasons.

**To Reproduce**
1. Open a PR.
2. Run the `reviewpad-cli` command on the opened PR (`./reviewpad-cli run -p [PULL_REQUEST] -t [GITHUB_TOKEN] -e [EVENT_PAYLOAD] -f [REVIEWPAD_FILE]`) on the branch of the PR.

**Expected behavior**
1. It should return the correct time, i.e, the creation time.

**Screenshots**
*None*

shay2025:
## Description

This pull request proposes a fix for the bug of the related issue.
The proposal is to directly retrieve the creation time of the issue when the timeline only has an event since we know this event is the creation of the issue.
This bug may be happening because the issue event types doesn't have the "opened" event - check https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types.

## Related issue

Closes #277 

## Type of change

## How was this tested?

It was done 2 tests.
1. Ran `task check -f`.
2. Opened a PR on the testing repo and ran the `reviewpad-cli` command on the branch of the PR: `./reviewpad-cli run -p [PULL_REQUEST] -t [GITHUB_TOKEN] -e [EVENT_PAYLOAD] -f [REVIEWPAD_FILE]`.

## Checklist

### DIFF
--- a/plugins/aladino/functions/lastEventAt.go
+++ b/plugins/aladino/functions/lastEventAt.go
@@ -43,7 +43,7 @@ func lastEventAtCode(e aladino.Env, args []aladino.Value) (aladino.Value, error)
 		return aladino.BuildIntValue(int(createdAtTime.Unix())), nil
 	}
 
-    lastEvent := timeline[len(timeline)-1]
+	lastEvent := timeline[len(timeline)-1]
 	var lastEventTime int
 	if *lastEvent.Event == "reviewed" {
 		lastEventTime = int(lastEvent.GetSubmittedAt().Unix())
--- a/plugins/aladino/functions/lastEventAt.go
+++ b/plugins/aladino/functions/lastEventAt.go
@@ -5,6 +5,8 @@
 package plugins_aladino_functions
 
 import (
+	"time"
+
 	"github.com/reviewpad/host-event-handler/handler"
 	"github.com/reviewpad/reviewpad/v3/lang/aladino"
 )
@@ -25,7 +27,23 @@ func lastEventAtCode(e aladino.Env, args []aladino.Value) (aladino.Value, error)
 		return nil, err
 	}
 
-	lastEvent := timeline[len(timeline)-1]
+	// The request for the issue timeline doesn't include the creation time of the first event.
+	// Since the first event is the opening of the issue, we know the last event time will be its creation time.
+	if len(timeline) == 1 {
+		createdAt, err := e.GetTarget().GetCreatedAt()
+		if err != nil {
+			return nil, err
+		}
+
+		createdAtTime, err := time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", createdAt)
+		if err != nil {
+			return nil, err
+		}
+
+		return aladino.BuildIntValue(int(createdAtTime.Unix())), nil
+	}
+
+    lastEvent := timeline[len(timeline)-1]
 	var lastEventTime int
 	if *lastEvent.Event == "reviewed" {
 		lastEventTime = int(lastEvent.GetSubmittedAt().Unix())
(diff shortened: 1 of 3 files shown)


marcelosousa:
We need more information as we can't reproduce right now the issue.

shay2025:
At the time of this comment, it seems Github API now returns an empty timeline upon the occurence of only one event which is the creation of the issue/pull request.

shay2025:
A workaround will be needed to take this in consideration.
The GraphQL API option (instead of REST API) is not taken into account because for the GQL query we would need to specify the parameters for each type of event and this would be too cumbersome since there are a lot of issue event types.

plugins/aladino/functions/lastEventAt_test.go
```diff
@@ -106,6 +106,24 @@ func TestLastEvent(t *testing.T) {
...
+						},
+					),
+				},
+				nil,
+				aladino.MockBuiltIns(),
+				nil,
+			),
+			wantVal: aladino.BuildIntValue(int(aladino.DefaultMockPrDate.Unix())),
+		},
+		// TODO: Add tests for when target entity is handler.Issue
```
shay2025, line 126: Context: issue #328

plugins/aladino/functions/lastEventAt.go
```diff
@@ -25,6 +27,26 @@ func lastEventAtCode(e aladino.Env, args []aladino.Value) (aladino.Value, error)
 		return nil, err
 	}
 
+	// If we are executing this built-in against an issue or a pr with no events (only the opening event), we have distinct behaviours.
+	// If it is an issue, we get an empty timeline.
+	// If it is a pr, we get one event (opening of the PR) with the majority of parameters as nil, including the createdAt which is the most relevant one in this context.
+	// This may be because the issue event types don't include the "opened" event.
+	// To verify the claim above, check: https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types.
+	// Since the first event is the opening of the issue/pr, we know the last event time will be its creation time.
+	if (entity.Kind == handler.PullRequest && len(timeline) == 1) || (entity.Kind == handler.Issue && len(timeline) == 0) {
```
marcelosousa, line 36: We need to separate the cases between the kinds.

marcelosousa (commented):
Please investigate why we cannot use the `updated_at` property instead of asking for the timeline events.

plugins/aladino/functions/lastEventAt.go
```diff
@@ -18,20 +20,15 @@ func LastEventAt() *aladino.BuiltInFunction {
...
 		return nil, err
 	}
 
-	lastEvent := timeline[len(timeline)-1]
-	var lastEventTime int
-	if *lastEvent.Event == "reviewed" {
-		lastEventTime = int(lastEvent.GetSubmittedAt().Unix())
-	} else {
-		lastEventTime = int(lastEvent.GetCreatedAt().Unix())
+	updatedAtTime, err := time.Parse("2006-01-02 15:04:05.999999999 -0700 MST", updatedAt)
```
ferreiratiago, line 28: I believe we can use `time.RFC3339` here.
```suggestion
	updatedAtTime, err := time.Parse(time.RFC3339 updatedAt)
```

ferreiratiago (changes requested):

ferreiratiago (approved):

PR #301, opened 2022-08-24
test: improve PaginatedRequest and ParseLink tests

shay2025:
## Description

This pull request introduces the improvement of PaginatedRequest and ParseLink related tests in order to adopt the TDT format; another improvement made was the adoption of `mock.WithRequestMatchPages` in order to mock paginated requests.
Besides this, it was removed the unit tests for getters since there is no logic to be tested there.

_Edit:_
It was disregarded the usage of `mock.WithRequestMatchPages` since this wouldn't help us cover the scenario of when a non first page request failed because we can't force an error for this request using the mentioned method.
Said that, we implement the pagination mock explicitly using `httpmock` for only 2 pages request.

## Related issue

*None*

## Type of change

## How was this tested?

Ran `task check -f` cmd.

## Checklist

### DIFF
--- a/codehost/github/pull_requests_test.go
+++ b/codehost/github/pull_requests_test.go
@@ -18,64 +18,6 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetPullRequestHeadOwnerName(t *testing.T) {
-	mockedHeadOwnerName := "reviewpad"
-	mockedPullRequest := aladino.GetDefaultMockPullRequestDetailsWith(&github.PullRequest{
-		Head: &github.PullRequestBranch{
-			Repo: &github.Repository{
-				Owner: &github.User{
-					Login: github.String(mockedHeadOwnerName),
-				},
-			},
-		},
-	})
-	wantOwnerName := mockedPullRequest.Head.Repo.Owner.GetLogin()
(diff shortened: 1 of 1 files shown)


codehost/github/pull_requests_test.go
```diff
@@ -19,192 +18,161 @@ import (
...
+						firstPageContent,
+						secondPageContent,
+					),
+				},
+				nil,
+				aladino.MockBuiltIns(),
+				nil,
+			),
+			wantVal: allPagesContent,
+		},
```
marcelosousa, line 100: Please add a unit test so that we can test when the request for the second page fails. This way we will be able to reach https://github.com/reviewpad/reviewpad/blob/main/codehost/github/pagination.go#L31

marcelosousa (changes requested):
Thanks for this PR 🙌  I've left a comment in the diff - I don't see any changes about ParseLink - maybe you can leave a comment in the code about that?

codehost/github/pull_requests_test.go
shay2025: @marcelosousa the parse link improvements are related with `parseNumPagesFromLink` function where the previous existing tests were converted to TDT which are contained in `TestParseNumPagesFromLink` in this PR.

marcelosousa (approved):

PR #315, opened 2022-08-26
feat: close issue/pull request with comment

### LINKED ISSUE #296: Close issue/pull request with comment
Right now, our `$close()` built-in only closes an issue/pull request. However, we may want to close it with comment.

Add string arg to the `$close()` built-in, being this new arg the close comment.

*None*

**Additional context**
*None*

shay2025:
## Description

This pull request introduces an improvement `$close()` by adding the possibility of closing an issue/pull request with a comment.

## Related issue

Closes #296

## Type of change

## How was this tested?

Ran `task check -f` cmd and used the `reviewpad-cli run` cmd against a testing repo with `reviewpad.yml` where a triggered rule had the `$close()` action.

## Checklist

### DIFF
--- a/codehost/github/target/issue_target.go
+++ b/codehost/github/target/issue_target.go
@@ -39,13 +39,20 @@ func (t *IssueTarget) GetNodeID() string {
 	return t.issue.GetNodeID()
 }
 
-func (t *IssueTarget) Close() error {
+func (t *IssueTarget) Close(comment string) error {
 	ctx := t.ctx
 	targetEntity := t.targetEntity
 	owner := targetEntity.Owner
 	repo := targetEntity.Repo
 	number := targetEntity.Number
 	issue := t.issue
+
+	if comment != "" {
+		if err := t.Comment(comment); err != nil {
+			return err
+		}
+	}
+
 	issue.State = github.String("closed")
 	issueRequest := &github.IssueRequest{
 		State: issue.State,
--- a/codehost/github/target/issue_target.go
+++ b/codehost/github/target/issue_target.go
@@ -40,19 +40,18 @@ func (t *IssueTarget) GetNodeID() string {
 }
 
 func (t *IssueTarget) Close(comment string) error {
-	ctx := t.ctx
-	targetEntity := t.targetEntity
-	owner := targetEntity.Owner
-	repo := targetEntity.Repo
-	number := targetEntity.Number
-	issue := t.issue
-
 	if comment != "" {
 		if err := t.Comment(comment); err != nil {
 			return err
 		}
 	}
 
+	ctx := t.ctx
+	targetEntity := t.targetEntity
+	owner := targetEntity.Owner
+	repo := targetEntity.Repo
+	number := targetEntity.Number
+	issue := t.issue
 	issue.State = github.String("closed")
 	issueRequest := &github.IssueRequest{
 		State: issue.State,
--- a/plugins/aladino/actions/close.go
+++ b/plugins/aladino/actions/close.go
@@ -11,12 +11,15 @@ import (
 
 func Close() *aladino.BuiltInAction {
 	return &aladino.BuiltInAction{
-		Type:           aladino.BuildFunctionType([]aladino.Type{}, nil),
+		Type:           aladino.BuildFunctionType([]aladino.Type{aladino.BuildStringType()}, nil),
 		Code:           closeCode,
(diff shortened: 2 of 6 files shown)


shay2025:
This https://github.com/reviewpad/docs/pull/20 needs to be merged before this PR is.

codehost/github/target/issue_target.go
```diff
@@ -39,7 +39,13 @@ func (t *IssueTarget) GetNodeID() string {
 	return t.issue.GetNodeID()
 }
 
-func (t *IssueTarget) Close() error {
+func (t *IssueTarget) Close(comment string) error {
+	if comment != "" {
```
marcelosousa, line 43: I think it makes more sense for this validation to happen the `t.Comment` function.

marcelosousa: It would actually be great to document in the Comment function what happens if the comment the empty string.

plugins/aladino/actions/close.go
```diff
@@ -11,12 +11,15 @@ import (
...
 		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest, handler.Issue},
 	}
 }
 
 func closeCode(e aladino.Env, args []aladino.Value) error {
-	return e.GetTarget().Close()
+	t := e.GetTarget()
+	comment := args[0].(*aladino.StringValue).Val
+
+	return t.Close(comment)
```
marcelosousa, line 24:
```suggestion
	return e.GetTarget().Close(comment)
```

plugins/aladino/actions/close_test.go
```diff
@@ -19,59 +19,123 @@ import (
...
+			assert.Equal(t, test.wantState, gotState)
+			assert.Equal(t, test.inputComment, closeComment)
+			assert.Equal(t, test.wantComment, commentCreated)
 
-	assert.Nil(t, err)
-	assert.Equal(t, wantState, gotState)
+			// Since these are variables common to all tests we need to reset their values at the end of each test
+			gotState = ""
+			closeComment = ""
+			commentCreated = false
```
marcelosousa, line 138: I don't think this is thread safe because if we run the tests in parallel we might get data races. Need to find another way to get the state :-)

marcelosousa (changes requested):

codehost/github/target/issue_target.go
shay2025: Idk if that would be desirable because whenever we comment `""` this action would be ignored, following your suggestion.
But, for now, we only want this to happen in the particular case of `Close()` in other cases we may want to throw an error on  other comments like `""` like the already defined default behavior. For example, `$comment("")` should return an error, should not?

plugins/aladino/actions/close_test.go
shay2025: I don't think the tests will be run in parallel.
Tests inside one `_test.go` file do not run in parallel by default - to enable parallel execution we would have to call `t.Parallel()` in the test case.

codehost/github/target/issue_target.go
marcelosousa: I think we need to check what exactly happens if a user tries $comment(""). I don't think it should return an error in that case.

shay2025: I took the decision of implementing that way because github throws an error saying that a comment can't have an empty body

shay2025: @marcelosousa I tried executing `$comment("")` in the PR https://github.com/explore-dev/didactic-robot/pull/68 and as expected a fail happens.
Log message: `Error: error running reviewpad team edition. Details POST https://api.github.com/repos/explore-dev/didactic-robot/issues/68/comments: 422 Validation Failed [{Resource:IssueComment Field:data Code:unprocessable Message:Body cannot be blank}]`

shay2025: I still believe `t.Comment()` should remain unchanged since this function has a direct purpose: create a comment.
If it doesn't have the conditions to do so (e.g., empty comment) it should fail - a comment CANNOT have a blank comment or else it fails to be a comment.
Said that, the validation you mentioned on the first comment of this thread, only makes sense to be made in `t.Close()` since this function can have two behaviours:
1. Perform a close WITH comment
2. Perform a close WITHOUT comment
It is a particular case.

marcelosousa (approved):

PR #330, opened 2022-08-31
fix: string parsing

### LINKED ISSUE #321: Investigate shift/reduce conflicts
**Describe the bug**
At the moment, when running the `task check -f` cmd, although everything runs smoothly, one of the messages is `conflicts: 6 shift/reduce`.
This occurence should be investigated this occurence.

**To Reproduce**
Run task check -f cmd.

**Expected behavior**
No shift/reduce conflicts.

**Screenshots**
image

ferreiratiago:
## Description

This pull request introduces a fix to string parsing in order to allow strings within strings.

The idea is basic and is to allow any character within a string, i.e. ".*".

However, in order for this to work properly we need to make sure that string are the last values to be parsed - therefore the order for expressions parsing has been updated.

This is a first proposal. Please provide your feedback 🙏

It also fixes the issues #321.

This issue was being cases by the problem  [Dangling else](https://en.wikipedia.org/wiki/Dangling_else).

You can check more details here: https://stackoverflow.com/a/1737613

## Related issue

Closes #321

## Type of change

## How was this tested?

Run `task check -f` without issues.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ reviewpad-cli
 .vscode
 .idea
 .DS_Store
+
+# Temporary files
+*.tmp
+temp/
--- a/lang/aladino/lex.go
+++ b/lang/aladino/lex.go
@@ -73,7 +73,7 @@ var tokens = []tokenDef{
 		token: IDENTIFIER,
 	},
 	{
-		regex: regexp.MustCompile(`^"[^"]*"`),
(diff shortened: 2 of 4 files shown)


ferreiratiago:
The solution of updating the regex for string values (i.e. ``^".*"``) does not work since the lexer runs before the parser.

PR #404, opened 2022-10-17
feat: set conditions for a review to be done with review built-in

### LINKED ISSUE #398: Set conditions to perform an automatic review ($review built-in)
The `$review` built-in performs a review independently if there were changes or not to the pull request. This shouldn't happen.

Only perform a review if there were changes to the pull request since the last review made by the `reviewpad-bot` (user that performs the automatic review).

shay2025:
## Description

The `$review` built-in seemed to have some flaws being one of them the fact that it performed a review whenever the action ran independently if the pull request had updates or not and if the last review was an `APPROVED`.
This built-in adds the condition that a review can only be performed if the pull request has updates since the last review made by `reviewpad-bot` and if the last review by that user is not an `APPROVED`.

## Related issue

Closes #398 

## Type of change

## How was this tested?

Ran `task check -f` cmd.

## Checklist

### DIFF
--- a/plugins/aladino/actions/review.go
+++ b/plugins/aladino/actions/review.go
@@ -23,7 +23,13 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
-	botAccount := e.GetGithubBotAccount().GetLogin()
+
+	botAccount, _, err := e.GetGithubClient().GetClientREST().Users.Get(e.GetCtx(), "")
+	if err != nil {
+		return err
+	}
+
+	botLogin := botAccount.GetLogin()
 
 	reviewEvent, err := parseReviewEvent(args[0].(*aladino.StringValue).Val)
 	if err != nil {
@@ -37,8 +43,8 @@ func reviewCode(e aladino.Env, args []aladino.Value) error {
 		return err
 	}
 
-	if codehost.HasReview(reviews, botAccount) {
-		lastReview := codehost.LastReview(reviews, botAccount)
+	if codehost.HasReview(reviews, botLogin) {
+		lastReview := codehost.LastReview(reviews, botLogin)
 
 		if lastReview.State == "APPROVED" {
 			return nil
--- a/plugins/aladino/actions/review.go
+++ b/plugins/aladino/actions/review.go
@@ -6,7 +6,6 @@ package plugins_aladino_actions
 
 import (
 	"fmt"
-	"log"
 
 	"github.com/reviewpad/reviewpad/v3/codehost"
 	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
@@ -24,9 +23,7 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
-	log.Printf("ACTION ACTOR2: %+v", *e.GetGithubActionActor())
 	actionActorLogin := e.GetGithubActionActor().GetLogin()
-	log.Printf("%+v", actionActorLogin)
 
 	reviewEvent, err := parseReviewEvent(args[0].(*aladino.StringValue).Val)
 	if err != nil {
@@ -47,7 +44,6 @@ func reviewCode(e aladino.Env, args []aladino.Value) error {
 			return nil
 		}
 
-		log.Printf("%+v", t.PullRequest.GetUpdatedAt())
 		if lastReview.SubmittedAt.After(t.PullRequest.GetUpdatedAt()) {
 			return nil
 		}
--- a/plugins/aladino/actions/review.go
+++ b/plugins/aladino/actions/review.go
@@ -24,6 +24,7 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
+	log.Printf("ACTION ACTOR2: %+v", *e.GetGithubActionActor())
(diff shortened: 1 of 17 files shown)


plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,39 @@ func Review() *aladino.BuiltInAction {
...
 		return err
 	}
 
-	reviewBody, err := checkReviewBody(reviewEvent, args[1].(*aladino.StringValue).Val)
+	reviewBody := args[1].(*aladino.StringValue).Val
+
+	// Only a review with APPROVE does not need a review comment
+	if reviewEvent != "APPROVE" && reviewBody == "" {
+		return fmt.Errorf("review: comment required in %v event", reviewEvent)
+	}
```
ferreiratiago, line 38: Why did we move this out of function? IMO, it was easier to understand with the function call `checkReviewBody`. WDYT?

plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,39 @@ func Review() *aladino.BuiltInAction {
...
+
+	reviews, err := t.GetReviews()
 	if err != nil {
 		return err
 	}
 
+	// If the last review of reviewpad-bot was an approval or if there were no updates to the pull request
+	// since last review made by reviewpad-bot, then do nothing.
```
ferreiratiago, line 46: No need for this comment.

plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,39 @@ func Review() *aladino.BuiltInAction {
...
+	if codehost.HasReview(reviews, reviewpadBot) {
+		lastReview := codehost.LastReview(reviews, reviewpadBot)
+
+		if lastReview.State == "APPROVED" {
+			return nil
+		}
+
+		if !lastReview.SubmittedAt.Before(t.PullRequest.GetUpdatedAt()) {
```
ferreiratiago, line 54:
```suggestion
		if lastReview.SubmittedAt.After(t.PullRequest.GetUpdatedAt()) {
```

Replace `! Before` with `After` for better reading.

ferreiratiago (approved):
Some minor suggestions. Please have a look.

plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,39 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
+	reviewpadBot := "reviewpad-bot"
```
ferreiratiago, line 26: Totally miss this. This is not ok. `reviewpad-bot` is a bot from reviewpad. Anyone else using this `review` built-in will either have `github-actions` or any other user. This needs to be rethink, otherwise it only works for reviewpad.

We need to check if the current user running reviewpad (github actions or any other) has done any review, and not `reviewpad-bot`.

ferreiratiago (changes requested):
Unfortunately this feature does not work as intended. It only works with reviewpad projects. Anyone outside of reviewpad won't be able to use it properly.

plugins/aladino/actions/review.go
shay2025: Should we use the workflow run actor?

shay2025: I have commited a change that follows that approach.

plugins/aladino/actions/review.go
```diff
@@ -21,17 +24,35 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
+	reviewer := e.GetGithubActionActor()
```
ferreiratiago, line 27: IMO we need to rename this variable to `actionActorLogin`. Being a `reviewer` is just a detail.

plugins/aladino/actions/review.go
```diff
@@ -21,17 +24,35 @@ func Review() *aladino.BuiltInAction {
...
 	}
 
+	if codehost.HasReview(reviews, reviewer) {
+		lastReview := codehost.LastReview(reviews, reviewer)
+
+		if lastReview.State == "APPROVED" {
+			return nil
+		}
+
+		log.Printf("%+v", t.PullRequest.GetUpdatedAt())
```
ferreiratiago, line 50: If we are keeping this log we need to make clear what it is. The way it is we will see a date but is not clear that is that is about.

```suggestion
		log.Printf("Pull request updated at: %+v", t.PullRequest.GetUpdatedAt())
```

ferreiratiago (changes requested):

plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,44 @@ func Review() *aladino.BuiltInAction {
 
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
+	entity := e.GetTarget().GetTargetEntity()
+
+	botAccount, _, err := e.GetGithubClient().GetClientREST().Users.Get(e.GetCtx(), "")
```
ferreiratiago, line 28: Why the name `botAccount`? This request provides the `authenticatedUser`

plugins/aladino/actions/review.go
```diff
@@ -21,17 +23,44 @@ func Review() *aladino.BuiltInAction {
...
+	if err != nil {
+		return err
+	}
+
+	botLogin := botAccount.GetLogin()
```
ferreiratiago, line 33:
```suggestion
	authenticatedUserLogin := authenticatedUser.GetLogin()
```

ferreiratiago (approved):
Minor suggestions.

plugins/aladino/actions/review.go
shay2025: Because the objective is to get the bot account because it is the authenticated user.

PR #405, opened 2022-10-17
feat: delete head branch built in action

### LINKED ISSUE #385: Automatic merged branch deletion built-in
On GitHub we can set for a repository to delete a head branch once a pull request is merged. However, we may only want this to happen only on certain cases.

It might come in handy to have a built-in `$deleteHeadBranch()` that deletes the head branch.

Use case:
```yaml
workflows:
  - name: delete-head-branch-after-merge
    description: Merge and delete head branch if patch only contains changes to files with extension .md
    if:
      - rule: '$hasFileExtensions([".md"])'
    then:
      - '$merge()'
      - '$deleteHeadBranch()'

```

zolamk:
## Description

This PR adds a new builtin action `$deleteHeadBranch()`

## Related issue

Closes #385 

## Type of change

## How was this tested?

Manual and unit tests.

## Checklist

- [X] I have performed a self-review of my code
- [X] I have added tests that prove my fix is effective or that my feature works 
- [X] I have ran `task check -f` and have no issues

- [X] Ask: this pull request requires a code review before merge

### DIFF
--- a/plugins/aladino/actions/deleteHeadBranch.go
+++ b/plugins/aladino/actions/deleteHeadBranch.go
@@ -1,0 +1,26 @@
+// Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+// Use of this source code is governed by a license that can be
+// found in the LICENSE file.
+
+package plugins_aladino_actions
+
+import (
+	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
+	"github.com/reviewpad/reviewpad/v3/handler"
+	"github.com/reviewpad/reviewpad/v3/lang/aladino"
+)
+
+func DeleteHeadBranch() *aladino.BuiltInAction {
+	return &aladino.BuiltInAction{
+		Type:           aladino.BuildFunctionType([]aladino.Type{}, nil),
+		Code:           deleteHeadBranch,
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func deleteHeadBranch(e aladino.Env, args []aladino.Value) error {
+	target := e.GetTarget().(*target.PullRequestTarget)
+	targetEntity := target.GetTargetEntity()
+
+	return e.GetGithubClient().DeleteReference(e.GetCtx(), targetEntity.Owner, targetEntity.Repo, "heads/"+*target.PullRequest.Head.Ref)
+}
--- a/plugins/aladino/actions/deleteHeadBranch.go
+++ b/plugins/aladino/actions/deleteHeadBranch.go
@@ -5,8 +5,6 @@
 package plugins_aladino_actions
 
 import (
-	"errors"
-
 	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
 	"github.com/reviewpad/reviewpad/v3/handler"
 	"github.com/reviewpad/reviewpad/v3/lang/aladino"
@@ -25,7 +23,7 @@ func deleteHeadBranch(e aladino.Env, args []aladino.Value) error {
 	targetEntity := target.GetTargetEntity()
 
 	if !*target.PullRequest.Merged && target.PullRequest.ClosedAt == nil {
-		return errors.New("pull request should be merged or closed before deleting head branch")
+		return nil
 	}
 
 	return e.GetGithubClient().DeleteReference(e.GetCtx(), targetEntity.Owner, targetEntity.Repo, "heads/"+*target.PullRequest.Head.Ref)
--- a/plugins/aladino/actions/deleteHeadBranch.go
+++ b/plugins/aladino/actions/deleteHeadBranch.go
@@ -5,6 +5,8 @@
 package plugins_aladino_actions
 
 import (
+	"errors"
+
 	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
 	"github.com/reviewpad/reviewpad/v3/handler"
 	"github.com/reviewpad/reviewpad/v3/lang/aladino"
@@ -21,6 +23,10 @@ func DeleteHeadBranch() *aladino.BuiltInAction {
 func deleteHeadBranch(e aladino.Env, args []aladino.Value) error {
 	target := e.GetTarget().(*target.PullRequestTarget)
 	targetEntity := target.GetTargetEntity()
+
+	if !*target.PullRequest.Merged && target.PullRequest.ClosedAt == nil {
+		return errors.New("pull request should be merged or closed before deleting head branch")
+	}
 
 	return e.GetGithubClient().DeleteReference(e.GetCtx(), targetEntity.Owner, targetEntity.Repo, "heads/"+*target.PullRequest.Head.Ref)
 }
--- a/plugins/aladino/actions/deleteHeadBranch_test.go
+++ b/plugins/aladino/actions/deleteHeadBranch_test.go
@@ -22,14 +22,14 @@ var deleteHeadBranch = plugins_aladino.PluginBuiltIns().Actions["deleteHeadBranc
 func TestDeleteHeadBranch(t *testing.T) {
 	now := time.Now()
 	isDeleteHeadBranchRequestPerformed := false
-	testCases := []struct {
+	tests := []struct {
 		name                    string
 		clientOptions           []mock.MockBackendOption
 		deleteShouldBePerformed bool
 		err                     error
 	}{
 		{
-			name: "success: pull request is closed",
+			name: "when pull request is closed",
 			clientOptions: []mock.MockBackendOption{
 				mock.WithRequestMatchHandler(
 					mock.GetReposPullsByOwnerByRepoByPullNumber,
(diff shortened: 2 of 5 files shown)


shay2025:
@zolamk I believe this action should only happen if and when the pull request is closed or merged. Otherwise, when the actions runs it just closes the pull requets and deletes the head branch. Could you add those conditions to the implementation, please? Thanks!

zolamk:
> @zolamk I believe this action should only happen if and when the pull request is closed or merged. Otherwise, when the actions runs it just closes the pull requets and deletes the head branch. Could you add those conditions to the implementation, please? Thanks!

I've updated the PR, but maybe we could also have built-in functions such as `$merged()` and `$closed()` we can use in a rule like this
```
workflows:
  - name: delete-closed-pr-head
    description: Delete Closed PR Head
    always-run: true
    if:
      - rule: '$closed() || $merged()'
        extra-actions:
          - "$deleteHeadBranch()"
```

marcelosousa:
@zolamk @shay2025 we need to investigate why the `Ask` label is not being added here.

zolamk:
> @zolamk @shay2025 we need to investigate why the `Ask` label is not being added here.

@marcelosousa I think that probably happened because I used a capital `X` to mark the ask option when creating the PR, in the reviewpad file for this repo uses a small letter `x` to check for the ask mode here https://github.com/reviewpad/reviewpad/blob/c9152a079bd2c8d5c0949e43fe19ab0e602d278b/reviewpad.yml#L94
perhaps it can be changed like `spec: '$contains($description(), "[x] Ask:") || $contains($description(), "[X] Ask:")'` or a case-insensitive version of `$contains` built-in function can be added

plugins/aladino/actions/deleteHeadBranch.go
```diff
@@ -0,0 +1,32 @@
...
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func deleteHeadBranch(e aladino.Env, args []aladino.Value) error {
+	target := e.GetTarget().(*target.PullRequestTarget)
+	targetEntity := target.GetTargetEntity()
+
+	if !*target.PullRequest.Merged && target.PullRequest.ClosedAt == nil {
+		return errors.New("pull request should be merged or closed before deleting head branch")
```
shay2025, line 28: Let's just return `nil` instead of an error.
This is because if the action runs in all pull requests if it comes across closed/merged ones the action will fail and it shouldn't happen.

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,140 @@
...
+		aladino.MockBuiltIns(),
+		nil,
+	)
+
+	err := deleteHeadBranch(mockedEnv, []aladino.Value{})
+
+	assert.NotNil(t, err)
+}
+
+func TestDeleteHeadBranch_WhenPRIsNotMergedOrClosed(t *testing.T) {
```
shay2025, line 52: Here you are testing the case scenario where the pull request is not merged AND closed.
We should do the following tests:
- when the pull request is closed but not merged;
- when the pull request is merged and closed (a merged pull request implies that it is closed but the other way around does not apply).

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,140 @@
...
+		[]mock.MockBackendOption{
+			mock.WithRequestMatchHandler(
+				mock.GetReposPullsByOwnerByRepoByPullNumber,
+				http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+					w.Write(mock.MustMarshal(mockedPullRequest))
+				}),
+			),
+			mock.WithRequestMatchHandler(
+				mock.DeleteReposGitRefsByOwnerByRepoByRef,
+				http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
shay2025, line 95: In order to check if the request to delete the head branch was performed, we could add a flag that indicated if the request was performed.
```suggestion
func TestDeleteHeadBranch_SuccessMerged(t *testing.T) {
    var isDeleteHeadBranchRequestPerformed bool
	mockedPullRequest := aladino.GetDefaultMockPullRequestDetailsWith(&github.PullRequest{
		Merged:   github.Bool(true),
		ClosedAt: nil,
	})
	mockedEnv := aladino.MockDefaultEnv(
		t,
		[]mock.MockBackendOption{
			mock.WithRequestMatchHandler(
				mock.GetReposPullsByOwnerByRepoByPullNumber,
				http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
					w.Write(mock.MustMarshal(mockedPullRequest))
				}),
			),
			mock.WithRequestMatchHandler(
				mock.DeleteReposGitRefsByOwnerByRepoByRef,
				http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
				   isDeleteHeadBranchRequestPerformed = true
				   w.WriteHeader(http.StatusNoContent)
				}),
			),
		},
		nil,
		aladino.MockBuiltIns(),
		nil,
	)

	err := deleteHeadBranch(mockedEnv, []aladino.Value{})

	assert.Nil(t, err)
	assert.True(t, isDeleteHeadBranchRequestPerformed, "the head branch should have been deleted")
}
```

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,140 @@
...
+		aladino.MockBuiltIns(),
+		nil,
+	)
+
+	err := deleteHeadBranch(mockedEnv, []aladino.Value{})
+
+	assert.Nil(t, err)
+}
+
+func TestDeleteHeadBranch_SuccessClosed(t *testing.T) {
```
shay2025, line 110: The same comment to the `TestDeleteHeadBranch_SuccessMerged` related to the `isDeleteHeadBranchRequestPerformed` is applicable here.

shay2025, line 110: This test function and the `TestDeleteHeadBranch_SuccessMerged` seem very similar only varying the `mockedPullRequest`. What about we apply Table-Driven tests here?
We could also include the `TestDeleteHeadBranch_WhenPRIsNotMergedOrClosed` test function as well.
If this seems a bit complex, let me know :slightly_smiling_face:

shay2025 (changes requested):
Just some minor changes

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+	mockMergedAndClosedPR := aladino.GetDefaultMockPullRequestDetailsWith(&github.PullRequest{
+		Merged:   github.Bool(true),
+		ClosedAt: &now,
+	})
+	mockNotMergedOrClosedPR := aladino.GetDefaultMockPullRequestDetailsWith(&github.PullRequest{
+		Merged:   github.Bool(false),
+		ClosedAt: nil,
+	})
+	isDeleteHeadBranchRequestPerformed := false
```
shay2025, line 40: This variable should be declared inside each test case like:
```
	for _, testCase := range testCases {
		t.Run(testCase.name, func(t *testing.T) {
              isDeleteHeadBranchRequestPerformed := false
...
```
This because the variable should not be shared between tests.

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/google/go-github/v45/github"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	"github.com/reviewpad/reviewpad/v3/lang/aladino"
+	plugins_aladino "github.com/reviewpad/reviewpad/v3/plugins/aladino"
+	"github.com/stretchr/testify/assert"
+)
+
+var deleteHeadBranch = plugins_aladino.PluginBuiltIns().Actions["deleteHeadBranch"].Code
+
+func TestDeleteHeadBranch(t *testing.T) {
```
shay2025, line 22: There is no need to declare so many mock pull requests.
Instead of separating the mock pr handlers we could pass a `clientOptions []mock.BackendOption` and pass that instead and this may avoid declaring so many mock pull requests since they can be passed directly to the appropriate handlers.

shay2025 (changes requested):

plugins/aladino/actions/deleteHeadBranch_test.go
shay2025: This can stay as it is for now. Check the other comments please :slightly_smiling_face:

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,193 @@
...
+				aladino.MockBuiltIns(),
+				nil,
+			)
+
+			err := deleteHeadBranch(mockedEnv, []aladino.Value{})
+
+			// this allows simplified checking of github error response equality
+			if e, ok := err.(*github.ErrorResponse); ok {
+				e.Response = nil
+			}
```
shay2025, line 184: Instead of adopting a strategy like this, you can follow what we had done in previous handling of such cases where instead of providing `err error` in the tests you can provide the error message and do:
```
gotErr := deleteHeadBranch(mockedEnv, []aladino.Value{})

if gotErr != nil && gotErr.(*github.ErrorResponse).Message != test.wantErr {
	assert.FailNow(t, "DeleteHeadBranch() error = %v, wantErr %v", gotErr, test.wantErr)
}
```

shay2025 (changes requested):

plugins/aladino/actions/deleteHeadBranch_test.go
zolamk: What happens in test scenarios where you are expecting an error, but an error isn't returned 🤔, i don't think this would work in those scenarios @shay2025 , what do you think?

shay2025: Yes, you are correct.

plugins/aladino/actions/deleteHeadBranch_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+var deleteHeadBranch = plugins_aladino.PluginBuiltIns().Actions["deleteHeadBranch"].Code
+
+func TestDeleteHeadBranch(t *testing.T) {
+	now := time.Now()
+	isDeleteHeadBranchRequestPerformed := false
+	tests := []struct {
+		name                    string
```
shay2025, line 26: You could do instead:
```
tests := map[string]struct {
		clientOptions           []mock.MockBackendOption
		deleteShouldBePerformed bool
		err                     error
}{
       {
			"when pull request is closed": {
			        clientOptions: []mock.MockBackendOption{
				    mock.WithRequestMatchHandler(
					       mock.GetReposPullsByOwnerByRepoByPullNumber,
...

for name, test := range tests {
	t.Run(name, func(t *testing.T) {
...
```

shay2025 (changes requested):
Almost there :slightly_smiling_face:

shay2025 (approved):

PR #414, opened 2022-10-24
feat: assign reviewers command

### LINKED ISSUE #408: Assign reviewers based on comment
The ability to quickly create a CLI to interact with Reviewpad could be useful to speed up certain processes.

This is particularly important in pull requests in open source projects where developers do not have the ability to do many actions.

As an example, they cannot assign reviewers to their pull requests. In such case, they could do so via a comment:

`/reviewpad assign reviewers` 

To trigger the process of finding reviewers.

This feature will likely also be necessary to add a pull request to the merge queue.

zolamk:
## Description
This PR allows assigning reviewers to a PR by adding a comment to it. Below is the format the command should take

`/reviewpad assign-reviewers <comma_separated_list_of_reviewers> <total_required_reviewers> <review_policy>`

### Example
`/reviewpad assign-reviewers marcelosousa, shay2025, ferreiratiago 2 random`

the command can also be used by just passing a list of reviewers as

`/reviewpad assign-reviewers marcelosousa, shay2025, ferreiratiago`

the review strategy can also be omitted while providing total required reviewers

`/reviewpad assign-reviewers marcelosousa, shay2025, ferreiratiago 2`

## Related issue

Closes #408 

## Type of change

## How was this tested?
Unit tests

## Checklist


shay2025 (changes requested):
Would it make sense to include the commands in their own package? We will have a lot of commands that will be implemented so the file `commands.go` would be too big.
Also, when a command is made through the comment, the action is as expected triggered but it runs the whole `reviewpad.yml` file again. It would be nice if only the workflow related to the command is run.

shay2025:
@zolamk upon testing, another thing happened, when executing `/reviewpad assign-reviewers ferreiratiago, shay2025` I received the following error: `strconv.Atoi: parsing "": invalid syntax`.

zolamk:
> Would it make sense to include the commands in their own package? We will have a lot of commands that will be implemented so the file `commands.go` would be too big...

Yeah, it makes sense to move it into its own package, but since we will get an import cycle error, we can add the available commands on the env, what do you think @shay2025?

shay2025:
> > Also, when a command is made through the comment, the action is as expected triggered but it runs the whole `reviewpad.yml` file again. It would be nice if only the workflow related to the command is run.
> 
> We can clear the rest of the workflows @shay2025, if we don't want them to run when a command is commented

Yes, that would be ideal but won't that affect the running of the original workflows when the action is triggered not by a command?

shay2025:
> > Would it make sense to include the commands in their own package? We will have a lot of commands that will be implemented so the file `commands.go` would be too big...
> 
> Yeah, it makes sense to move it into its own package, but since we will get an import cycle error, we can add the available commands on the env, what do you think @shay2025?

Yes, we could do that. But it still remains the situation that since we will have a lot of command, it makes sense that each command has its file and maybe the env can have a map of the command like it does with the built-ins.

zolamk:
> > > Also, when a command is made through the comment, the action is as expected triggered but it runs the whole `reviewpad.yml` file again. It would be nice if only the workflow related to the command is run.
> > 
> > 
> > We can clear the rest of the workflows @shay2025, if we don't want them to run when a command is commented
> 
> Yes, that would be ideal but won't that affect the running of the original workflows when the action is triggered not by a command?

We'll only be clearing them in case a command is activated, the rest of the time the workflows in the `reviewpad.yml` file will execute as normal

engine/exec.go
```diff
@@ -112,6 +112,33 @@ func Eval(file *ReviewpadFile, env *Env) (*Program, error) {
...
+					return nil, err
+				}
+
+				program.append(actions)
+
+				break
+			}
+		}
+	}
+
```
shay2025, line 141: Maybe we could try to move this code block to the beginning of the `Eval` func and at the end of the commands processing we could just return the program.
```
// a program is a list of statements to be executed based on the command, workflow rules and actions.
program := BuildProgram(make([]*Statement, 0))

// process commands
if env.TargetEntity.EventName == "issue_comment" {
  ...
  return program, nil
}
```
That way we only deal with the commands with no processing related to the `reviewpad.yml` file and we don't need to add:
```
file.Rules = []PadRule{}
file.Workflows = []PadWorkflow{}
```
in order to clear the rules and workflows.

shay2025 (changes requested):

engine/exec.go
zolamk: 🤔 I think we actually need to move it down, right after labels and groups have been processed because in the future we might have a command that relies on the groups defined in the files or something, what do you think?

shay2025: Yes, you're right :slightly_smiling_face:

engine/commands/assign_reviewer_command.go
```diff
@@ -0,0 +1,48 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var (
+	assignReviewerCommandRegex = regexp.MustCompile(`^\/reviewpad assign-reviewers\s+((?:[a-zA-Z0-9\-]+[a-zA-Z0-9])(?:,\s*[a-zA-Z0-9\-]+[a-zA-Z0-9])*)(?:\s+(\d+))?(?:\s+(random|round-robin|reviewpad))?$`)
+)
+
+func AssignReviewerCommand(matches []string) ([]string, error) {
```
shay2025, line 19: Since this is inserted in the commands package we know that we are dealing with `assign reviewers` in a command context so I believe there is no need to mention `Command`.
```suggestion
func AssignReviewer(matches []string) ([]string, error) {
```

The same applies to `assignReviewerCommandRegex` that can be `assignReviewerRegex`. And also the file names, which can be `assignReviewer.go` and `assignReviewer_test.go` instead of `assign_reviewer_command.go` and `assign_reviewer_command_test.go`, respectively.

shay2025 (changes requested):
Almost there :slightly_smiling_face:

shay2025:
@zolamk when you can please update the branch with the latest version in `main`.

shay2025 (approved):

PR #427, opened 2022-10-26
feat: comment metric report when pr is merged

### LINKED ISSUE #407: Get a metrics report when the pull request is merged
As soon as a pull request is merged, it is unclear what could have been done better. 
The ability to show some statistics could help gamify and reduce the time to merge.
Some of these stats could be:
- Coding Time: measures the time it takes from the first commit until a pull request is issued.
- Pickup Time: measures the time a pull request is issued until the first review.
- Review Time: measures the time from the first review until the pull request is merged.

The stats should be shown in a comment to the pull request as soon as it is closed.

zolamk:
## Description
This PR comments metrics reports when a pull request is merged.

## Related issue

Closes #407 

## Type of change

## Checklist


shay2025 (changes requested):
Why did you disregard the previous reviewpad report?
Both reports should exist and they are shown to the user if the `mode` property in the `reviewpad.yml` is set to `verbose`.

zolamk:
> Why did you disregard the previous reviewpad report? Both reports should exist and they are shown to the user if the `mode` property in the `reviewpad.yml` is set to `verbose`.

The reason I made it into a separate report is because the metrics should only be reported when the PR is merged but we can also combine the two reports into one and check if the PR is merged when performing the report, what do you think @shay2025 ?

shay2025:
@zolamk what is happening is that the previous report annotation does not seem to exist. These are two different reports. One with the reviewpad executions (the already existing one) and the metrics report. And what it seems to have been done is that the annotation that helped identify the first report was disregarded.
Both reports should have separared implementations, i.e, the implementation of the first report that already existed should remain untouched and a new report implementation should be done for the metrics report.
The only thing in common for both reports should be that they are only shown to the user if the `mode` property in the `reviewpad.yml` file is set  to `verbose`.

shay2025:
@zolamk the report doesn't seem to be printing the time data correctly. I opened a pull request that was merged 1 hour after it was opened and the `coding time` is incorrect.
image

zolamk:
> @zolamk the report doesn't seem to be printing the time data correctly. I opened a pull request that was merged 1 hour after it was opened and the `coding time` is incorrect. image

@shay2025 I thought coding time was calculated as the difference between the first commit and the time the pull request was created, or is it supposed to be the time the pull request was merged?

shay2025:
> > @zolamk the report doesn't seem to be printing the time data correctly. I opened a pull request that was merged 1 hour after it was opened and the `coding time` is incorrect. image
> 
> @shay2025 I thought coding time was calculated as the difference between the first commit and the time the pull request was created, or is it supposed to be the time the pull request was merged? I think that was how it was defined on the issue and that's how the people at linearb define it https://linearb.io/blog/reducing-coding-time/

The time shown should be precise, i.e, say the exact seconds/minutes/days I believe.

utils/time.go
```diff
@@ -0,0 +1,60 @@
...
+func pluralize(part int, label string) string {
+	if part > 1 {
+		return label + "s"
+	}
+
+	return label
+}
+
+func FormatTimeDiff(x time.Time, y time.Time) string {
+	years, months, days, hours, minutes, seconds := timex.Diff(x, y)
```
shay2025, line 24: Why are we using the `timex.Diff` method? We could use the `Sub` method from the `time` library:
```
years, months, days, hours, minutes, seconds := x.Sub(y)
```
That way we don't need to import another library.

utils/time.go
```diff
@@ -0,0 +1,60 @@
...
+
+	l := len(parts)
+
+	if l == 1 {
+		return parts[0]
+	} else if l == 2 {
+		return parts[0] + ", and " + parts[1]
+	}
+
+	return strings.Join(parts[:l-1], ", ") + ", and " + parts[l-1]
```
shay2025, line 59: There is no need for the comma in ", and".

utils/time.go
```diff
@@ -0,0 +1,60 @@
...
+	if seconds > 0 {
+		parts = append(parts, fmt.Sprintf("%d %s", seconds, pluralize(seconds, "second")))
+	}
+
+	l := len(parts)
+
+	if l == 1 {
+		return parts[0]
+	} else if l == 2 {
+		return parts[0] + ", and " + parts[1]
```
shay2025, line 56: There is no need for the comma in ", and".

shay2025 (changes requested):

utils/time.go
zolamk: The thing is [Sub](https://pkg.go.dev/time#Time.Sub) gives you a `Duration`, it doesn't divide the diff into it's `year`, `month`, `days`, `hours`, `minutes`, `seconds` parts but yeah it might not be justifiable to include a new dependency for one function, if you'd prefer we can just implement the diff ourselves, what do you think?

shay2025: Yes, it would be preferable to have our own method like `func diff(a, b time.Time) (year, month, day, hour, min, sec int)`.

runner.go
```diff
@@ -85,6 +85,14 @@ func Run(
 		}
 	}
 
+	if targetEntity.Kind == handler.PullRequest && targetEntity.EventName == "pull_request" && targetEntity.EventAction == "closed" {
```
ferreiratiago, line 88: Can we add this condition into a single function and call it `isPullRequestReadyForReportMetrics`.

utils/time.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if month < 0 {
+		month += 12
+		year--
+	}
+
+	return
+}
+
+func FormatTimeDiff(x time.Time, y time.Time) string {
```
ferreiratiago, line 77:
```suggestion
func ReadableTimeDiff(x time.Time, y time.Time) string {
```

ferreiratiago (changes requested):

codehost/github/pull_requests.go
```diff
@@ -406,3 +432,43 @@ func (c *GithubClient) DeleteReference(ctx context.Context, owner, repo, ref str
...
+	err := c.GetClientGraphQL().Query(ctx, &firstReviewDateQuery, varGQLFirstReviewDate)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(firstReviewDateQuery.Repository.PullRequest.Reviews.Nodes) == 1 {
+		return &firstReviewDateQuery.Repository.PullRequest.Reviews.Nodes[0].CreatedAt, nil
+	}
+
+	return nil, nil
```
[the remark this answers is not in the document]
zolamk: Yeah makes sense, should be faster as well because we can combine the GraphQL queries

lang/aladino/interpreter.go
```diff
@@ -195,6 +198,65 @@ func (i *Interpreter) Report(mode string, safeMode bool) error {
...
+		report.WriteString(fmt.Sprintf("\n**👀 Review Time**: %s", utils.FormatTimeDiff(*pr.MergedAt, *firstReviewDate)))
+	}
+
+	if report.Len() > 0 {
+		comment, err := FindReportCommentByAnnotation(i.Env, ReviewpadMetricReportCommentAnnotation)
+		if err != nil {
+			return err
+		}
+
+		r := ReviewpadMetricReportCommentAnnotation + "\n## 📈 Pull Request Metrics\n" + report.String()
```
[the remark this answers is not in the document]
zolamk: The thing is the markdown won't be treated as a header unless we have that space there, I think we need to keep the space

ferreiratiago:
@zolamk can you please rebase the pull request with the base branch? 🙏

zolamk:
> @zolamk can you please rebase the pull request with the base branch? pray

sure thing, on it.

zolamk:
> @zolamk can you please rebase the pull request with the base branch? pray

done @ferreiratiago

lang/aladino/interpreter.go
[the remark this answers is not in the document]
ferreiratiago: You're right 👍

ferreiratiago:
## 📈 Pull Request Metrics
**💻 Coding Time**: 3 minutes and 9 seconds
**🛻 Pickup Time**: 7 days, 16 hours, 27 minutes and 15 seconds
**👀 Review Time**: 5 days, 10 hours, 53 minutes and 56 seconds

PR #433, opened 2022-10-31
feat: improve assign reviewer command

### LINKED ISSUE #428: Improve assign-reviewers command
In general, a shell command consists of the command itself followed by arguments, separated by spaces.
The command can also have flags (with aliases) which present optional arguments.

Right now, the `assign-reviewers` command does not follow that convention.
The command allows for spaces in its first argument, which is the list of reviewers that can be assigned (e.g. `john, mary, peter`) which for now we won't support.
And the optional arguments (`total_reviewers` and `review_policy`) order is not guaranteed. The optional arguments should be flagged and the order of how they appear should not matter.

Besides this, the command is found via a regex which matches the full command (e.g. `/reviewpad assign-reviewers a,b,c 2 random`). 
Ideally, first Reviewpad should check for the command `/reviewpad assign-reviewers` which can indeed be done via regex; and only then process its arguments.

1. Reviewpad should first check for the command `/reviewpad assign-reviewers` which can indeed be done via regex; and then process its arguments.
If the arguments are wrongly inputed (e.g. wrong types) the action should not fail but log a warning exposing why the command was not executed.
  2. Add flags for the optional arguments where the order by how they appear does not matter:
- `-t/--total-reviewers`
- `-p/--review-policy`
Examples:
```
/reviewpad assign-reviewers john,peter -t 1 -p random
```
```
/reviewpad assign-reviewers john,peter --review-policy random -t 1
```

3. Do not allow spaces inside arguments - e.g. `jane, john, peter` is allowed at the moment but it should only be allowed `jane,john,peter`.

zolamk:
## Description
This PR improves the `assign-reviewers` command to streamline configuration passing
### Examples
- `/reviewpad assign-reviewers john,jane`
- `/reviewpad assign-reviewers john,jane --total-reviewers=1`
- `/reviewpad assign-reviewers jane,john --total-reviewers 1 --review-policy=round-robin`
- `/reviewpad assign-reviewers jane,john --review-policy random`
- `/reviewpad-assign-reviewers john,jane --review-policy random --total-reviewers 2`

List any dependencies that are required for this change (if applicable.)
- https://github.com/mattn/go-shellwords

## Related issue

Closes #428 

## Type of change

## How was this tested?
Manual & Unit Tests.

## Checklist


engine/commands/assignReviewer_test.go
```diff
@@ -15,82 +15,89 @@ import (
...
 				"john-123-jane",
-				"1",
-				"reviewpad",
+				"--total-reviewers=1",
+				"--review-policy=reviewpad",
 			},
-			wantActions: []string{`$assignReviewer(["john-123-jane"], 1, "reviewpad")`},
+			wantAction: `$assignReviewer(["john-123-jane"], 1, "reviewpad")`,
 		},
 		"when only two reviewers is provided": {
```
shay2025, line 71:
```
"when only two reviewers are provided"
```

engine/commands/assignReviewer_test.go
```diff
@@ -15,82 +15,89 @@ import (
...
-				"reviewpad",
+			args: []string{
+				"jane,john",
+				"--total-reviewers=1",
+				"--review-policy=round-robin",
 			},
-			wantActions: []string{`$assignReviewer(["jane","john"], 1, "reviewpad")`},
+			wantAction: `$assignReviewer(["jane","john"], 1, "round-robin")`,
 		},
 	}
```
shay2025, line 87: We should also add the test case: `"when number of provided reviewers is less than requested reviewers"`

engine/exec.go
```diff
@@ -126,21 +129,37 @@ func Eval(file *ReviewpadFile, env *Env) (*Program, error) {
...
-			matches := r.FindAllStringSubmatch(env.TargetEntity.Comment, 1)
+	if env.TargetEntity.EventName == "issue_comment" &&
+		env.TargetEntity.EventAction == "created" &&
+		env.TargetEntity.Comment.Body != nil &&
+		strings.HasPrefix(*env.TargetEntity.Comment.Body, "/reviewpad") {
+		comment := *env.TargetEntity.Comment.Body
+		out := new(bytes.Buffer)
+		comment = strings.TrimPrefix(comment, "/reviewpad")
+
+		args, err := shellwords.Parse(comment)
```
shay2025, line 140: We should not need this extra library because the args should be separated by spaces I believe.

shay2025 (changes requested):

engine/commands/assignReviewer.go
```diff
@@ -8,39 +8,64 @@ import (
...
+		RunE: AssignReviewer,
 	}
 
-	reviewersList := strings.Split(strings.ReplaceAll(matches[1], " ", ""), ",")
+	flags := assignReviewerCmd.Flags()
 
-	availableReviewers := `"` + strings.Join(reviewersList, `","`) + `"`
+	flags.StringP("review-policy", "p", "reviewpad", "valid values are random, round-robin, reviewpad, default value is reviewpad")
+
+	flags.Int8P("total-reviewers", "t", 1, "Total number of reviewers, default value is 1")
```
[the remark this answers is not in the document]
zolamk: We are using `1` reviewer by default, should we make it all by default?

engine/commands/assignReviewer_test.go
zolamk: currently if the number of provider reviewers is less than the requested reviewers the command will still build an action and add it to the program

zolamk: should we have a validation in place for it?

engine/commands/assignReviewer.go
```diff
@@ -8,39 +8,64 @@ import (
...
+
+	"github.com/spf13/cobra"
 )
 
-var assignReviewerRegex = regexp.MustCompile(`^\/reviewpad assign-reviewers\s+((?:[a-zA-Z0-9\-]+[a-zA-Z0-9])(?:,\s*[a-zA-Z0-9\-]+[a-zA-Z0-9])*)(?:\s+(\d+))?(?:\s+(random|round-robin|reviewpad))?$`)
+func AssignReviewerCmd() *cobra.Command {
+	assignReviewerCmd := &cobra.Command{
+		Use:           "assign-reviewers",
+		Short:         "Assign Reviewer To Pull Request",
+		SilenceUsage:  true,
```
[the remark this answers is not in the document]
zolamk: It shouldn't since we are using `RunE` and handling the errors ourselves, any errors should be returned by `Execute` at which point we handle them by updating the comment with the error

engine/exec.go
zolamk: I think spaces inside of quotations and things like those need to be considered, but yeah we can use regexp to parse it

engine/commands/assignReviewer_test.go
shay2025: There is no need for the validation I believe. The `$assignReviewer` built-in when receives a list of reviewers whose size is less than the number of requested reviewers then it will consider the number of requested reviewers to be the size of the provided list of reviewers instead.

engine/exec.go
shay2025: I noticed the examples you provided in the PR description that the flags values are preceded by an `=`. Do the commands still work if it is something like `/reviewpad assign-reviewers john,jane --total-reviewers 1`? If not, we should consider to also support that.
If so, then we can leave the `shellwords` library I believe.

shay2025: This that library can still separate the arguments by the spaces usage.

engine/commands/assignReviewer.go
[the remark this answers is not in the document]
shay2025: Yes, because that's the behaviour of the `$assignReviewer` built-in: https://docs.reviewpad.com/guides/built-ins#assignreviewer

engine/exec.go
zolamk: Yeah, the library can handle those cases as well, so I believe we should keep it

shay2025 (approved):
After running some tests, the feature seems to be working as expected :slightly_smiling_face: 
There is only a minor strange behavior which is upon the execution of the command the existing report will be edited where the `execution actions` block is replaced with the command action. After the command execution finishes, the report is edited again with the previous content.
But this behavior can be dealt in a separate issue.


https://user-images.githubusercontent.com/61353903/200826099-ab6862f0-00d5-420f-be24-079ca92ceced.mp4

shay2025:
@zolamk when you can please resolve the conflicts :slightly_smiling_face:

ferreiratiago (changes requested):
The major change we need to do here is to not use `TargetEntity` to append all info not related to the entity but with the action to be applied to the entity.

handler/processors.go
```diff
@@ -36,7 +36,7 @@ type TargetEntity struct {
 	Owner       string
 	Repo        string
 	EventName   string
```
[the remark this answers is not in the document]
zolamk: `Kind` represents what the target entity is currently `pull request` or `issue`, `EventName` is the actual event that happened which could be things like `issue_comment`, `pull_request_review`, `discussion_comment`...

engine/exec.go
```diff
@@ -126,21 +129,37 @@ func Eval(file *ReviewpadFile, env *Env) (*Program, error) {
...
+			comment := fmt.Sprintf("%s\n```\n🚫 error\n\n%s\n```", *env.TargetEntity.Comment.Body, err.Error())
 
-				return program, nil
+			if _, _, err := env.GithubClient.EditComment(env.Ctx, env.TargetEntity.Owner, env.TargetEntity.Repo, *env.TargetEntity.Comment.ID, &github.IssueComment{
+				Body: github.String(comment),
+			}); err != nil {
+				return nil, err
 			}
+
+			return nil, err
```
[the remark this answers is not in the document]
zolamk: I haven't looked into how other bots handle it but, I think as far as user experience is concerned, editing the comment to show what went wrong is a good experience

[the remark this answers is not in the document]
zolamk: As for the user editing the comment, we are only processing commands on comment created event since we didn't want to get into some sort of loop where reviewpad keeps getting triggered by its own comment so the user editing their comment shouldn't matter, What do you think @ferreiratiago

handler/processors.go
```diff
@@ -36,7 +36,7 @@ type TargetEntity struct {
...
-	Comment     string
+	Comment     *github.IssueComment
```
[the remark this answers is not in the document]
zolamk: Yeah, make much better sense to have another structure for this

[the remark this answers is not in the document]
zolamk: We already have an `ActionEvent` should we reuse that or introduce a new, I say introduce a new because we might not need all the fields already available in the current `ActionEvent` what do you think @ferreiratiago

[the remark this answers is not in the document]
zolamk: @ferreiratiago I've submitted an update for this as a separate pr #459 , i'll rebase this after that's reviewed and merged

ferreiratiago (changes requested):
The pr #459 was merged. Can you please @zolamk update this pr? Thanks 🙏

ferreiratiago (approved):

PR #446, opened 2022-11-11
fix: don't re-request review unless a new commit has been pushed

### LINKED ISSUE #415: Re-requesting a review from a user with a updated review
**Describe the bug**
For some reason, reviewpad re-requests a review from a user with an updated review. By updated it means that the user did a review and no changes were added since.

**To Reproduce**
- Open a pr
- Reviewpad requests a review from user X
- User X makes a `request changes` review
- Reviewpad runs and re-requests a review

**Expected behavior**
- Reviewpad shouldn't re-request a review when the user has an updated review (regardless of review type - i.e. approved, requested changes, commented).

**Screenshots**
The problem can be seen in the following pr https://github.com/reviewpad/reviewpad/pull/412

Screen Shot 2022-10-24 at 8 49 03 PM

zolamk:
## Description
This PR updates the `$assignReviewer` built-in action so that it checks the last pushed date before re-requesting a review

## Related issue

Closes #415 

## Type of change

## How was this tested?
Unit tests and manual tests

## Checklist

### DIFF
--- a/plugins/aladino/actions/assignReviewer.go
+++ b/plugins/aladino/actions/assignReviewer.go
@@ -37,6 +37,11 @@ func assignReviewerCode(e aladino.Env, args []aladino.Value) error {
 	totalRequiredReviewers := args[1].(*aladino.IntValue).Val
 	policy := args[2].(*aladino.StringValue).Val
 	target := e.GetTarget().(*target.PullRequestTarget)
+	targetEntity := target.GetTargetEntity()
+	ctx := e.GetCtx()
+	owner := targetEntity.Owner
+	repo := targetEntity.Repo
+	prNum := targetEntity.Number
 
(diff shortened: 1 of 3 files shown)


plugins/aladino/actions/assignReviewer.go
```diff
@@ -71,12 +76,17 @@ func assignReviewerCode(e aladino.Env, args []aladino.Value) error {
 		return err
 	}
 
+	lastPushDate, err := e.GetGithubClient().GetPullRequestLastPushDate(ctx, owner, repo, prNum)
+	if err != nil {
+		return err
+	}
```
ferreiratiago, line 82: This is not the right way to do it. We should do this on `target` which already has the client in it.

Please have a look at `codehost/github/target/pull_request_target.go`.

ferreiratiago (approved):

Issue #435, opened 2022-11-03
Create builtin to extract check status


ferreiratiago:
**Is your feature request related to a problem? Please describe.**
Currently, reviewpad does not have the ability to extract a check status for [check run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#check_run) and [check suite](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#check_suite).

We need to add the ability to fetch the status for a check run and check suite.

**Describe the solution you'd like**
First, we need to understand the difference between a check run and a check suite. Then, evaluate if both built-ins are required and what are the possible status a check run/suite can return.

After getting all this info we need to implement the required builtins in order to fetch the check run/suite status.

Proposed solutions:
```yml
$checkRunStatus(<RUN_NAME>)
$checkSuiteStatus(<SUITE_NAME>)
```

shay2025:
@ferreiratiago lookint at the implementation of the built-in `$workflowStatus`, it seems to be the same as it would be for `$checkRunStatus(<RUN_NAME>)`, does it not?
https://github.com/reviewpad/reviewpad/blob/main/plugins/aladino/functions/workflowStatus.go

shay2025:
A check run can have one of the following status: `queued`, `in_progress`, `completed` (https://docs.github.com/en/rest/guides/getting-started-with-the-checks-api#about-check-runs).
While a check suite, which is a set of check runs, summarize the status and conclusion of the check runs that a suite includes.

shay2025:
It would make sense to have two built-ins one that checks the status of a `check run` and another that checks the status of a `check suite`. This because we may only want to know the status of an individual check since a check suite may contain more than one check run.

shay2025:
I don't know if it is possible to get a check suite by name since the github check suite object does not contain any field with the name https://docs.github.com/en/rest/checks/suites#get-a-check-suite

PR #451, opened 2022-11-15
feat: checkRunStatus built-in

shay2025:
## Description

This pull request introduces a new built-in function called `$checkRunStatus` that returns that status of the given check being that status can be one of the following: `queued`, `in_progress`, `completed` (https://docs.github.com/en/rest/guides/getting-started-with-the-checks-api#about-check-runs).

## Related issue

Related to #435 

## Type of change

## How was this tested?

Ran `task build -f` and `task test -f`.
Ran this reviewpad version against a dummy PR of a repo that had reviewpad installed.

## Checklist


plugins/aladino/functions/checkRunStatus.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func checkRunStatusCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
+	checkRunName := args[0].(*aladino.StringValue).Val
+	entity := e.GetTarget().GetTargetEntity()
+	owner := entity.Owner
+	repo := entity.Repo
+	number := entity.Number
+
+	commits, _, err := e.GetGithubClient().GetClientREST().PullRequests.ListCommits(e.GetCtx(), owner, repo, number, &github.ListOptions{})
```
ferreiratiago, line 28: This is not the correct way to do it. We should create a function within pull request target (`codehost/github/target/pull_request_target.go`), e.g. `GetCommits`. In fact, that function is already created. Please have a look.

plugins/aladino/functions/checkRunStatus.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	lastCommitSha := commits[len(commits)-1].GetSHA()
```
ferreiratiago, line 33: I don't think so this is the right way to do it. Does the API guarantee that the last commit is the most recent? I don't see that on the [API description](https://docs.github.com/en/rest/pulls/pulls#list-commits-on-a-pull-request). Also, this request is paginated, which means that it's not guaranteed that we are getting the last commit.

ferreiratiago (changes requested):

plugins/aladino/functions/checkRunStatus.go
shay2025: It might only not be sorted in a rebase scenario https://github.blog/changelog/2020-07-14-pull-request-commits-now-ordered-chronologically/

shay2025: But indeed, the pagination needs to be handled.

plugins/aladino/functions/checkRunStatus.go
shay2025: The  existing function `GetCommits` indeed exists but it doesn't return the object we want - without the SHA's commit. We should use the `GetPullRequestCommits` function that lives in `codehost/github/pull_request.go`.

ferreiratiago (approved):

PR #480, opened 2022-11-28
feat: add hasBinaryFile built-in function

### LINKED ISSUE #477: Add builtin hasBinaryFile
Create a new built-in to check if a change has a binary file.

This built-in is only valid for pull requests. It shouldn't have any action when running on issues.

Look for a binary file in the pull request patch.

zolamk:
## Description
This PR adds a new `hasBinaryFile` built-in function that returns a boolean value indicating whether a PR has binary file changes.

## Related issue

Closes #477 

## Type of change

## How was this tested?
Unit & Manual Tests

## Checklist

### DIFF
--- a/plugins/aladino/functions/hasBinaryFile.go
+++ b/plugins/aladino/functions/hasBinaryFile.go
@@ -20,15 +20,13 @@ func HasBinaryFile() *aladino.BuiltInFunction {
 
 func hasBinaryFile(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
 	patch := e.GetTarget().(*target.PullRequestTarget).Patch
-	hasBinaryFile := false
 
 	for _, patchFile := range patch {
 		// file is binary if GitHub does not display a textual diff
 		if patchFile.Repr.GetPatch() == "" {
-			hasBinaryFile = true
-			break
+			return aladino.BuildBoolValue(true), nil
 		}
 	}
 
-	return aladino.BuildBoolValue(hasBinaryFile), nil
+	return aladino.BuildBoolValue(false), nil
 }
--- a/plugins/aladino/functions/hasBinaryFile.go
+++ b/plugins/aladino/functions/hasBinaryFile.go
@@ -1,0 +1,34 @@
+// Copyright 2022 Explore.dev Unipessoal Lda. All Rights Reserved.
+// Use of this source code is governed by a license that can be
+// found in the LICENSE file.
+
+package plugins_aladino_functions
+
+import (
+	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
+	"github.com/reviewpad/reviewpad/v3/handler"
+	"github.com/reviewpad/reviewpad/v3/lang/aladino"
+)
+
+func HasBinaryFile() *aladino.BuiltInFunction {
+	return &aladino.BuiltInFunction{
+		Type:           aladino.BuildFunctionType([]aladino.Type{}, aladino.BuildBoolType()),
+		Code:           hasBinaryFile,
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func hasBinaryFile(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
+	patch := e.GetTarget().(*target.PullRequestTarget).Patch
+	hasBinaryFile := false
+
+	for _, patch := range patch {
+		// there won't be any patch for binary files, we can use that to check if pr has binary file changes
+		if patch.Repr.GetPatch() == "" {
+			hasBinaryFile = true
+			break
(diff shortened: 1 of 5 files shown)


plugins/aladino/functions/hasBinaryFile.go
```diff
@@ -0,0 +1,34 @@
...
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func hasBinaryFile(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
+	patch := e.GetTarget().(*target.PullRequestTarget).Patch
+	hasBinaryFile := false
+
+	for _, patch := range patch {
+		// there won't be any patch for binary files, we can use that to check if pr has binary file changes
```
marcelosousa, line 26:
```suggestion
		// file is binary if GitHub does not display a textual diff
```

plugins/aladino/functions/hasBinaryFile.go
```diff
@@ -0,0 +1,34 @@
...
+		Code:           hasBinaryFile,
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func hasBinaryFile(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
+	patch := e.GetTarget().(*target.PullRequestTarget).Patch
+	hasBinaryFile := false
+
+	for _, patch := range patch {
```
marcelosousa, line 25: Please do not use the same variable on the left side - it's very confusing. I suggest `patchFile`.

plugins/aladino/functions/hasBinaryFile.go
```diff
@@ -0,0 +1,34 @@
...
+		// there won't be any patch for binary files, we can use that to check if pr has binary file changes
+		if patch.Repr.GetPatch() == "" {
+			hasBinaryFile = true
+			break
```
marcelosousa, line 29: I think you can simply return here.

plugins/aladino/functions/hasBinaryFile.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func hasBinaryFile(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
+	patch := e.GetTarget().(*target.PullRequestTarget).Patch
+	hasBinaryFile := false
+
+	for _, patch := range patch {
+		// there won't be any patch for binary files, we can use that to check if pr has binary file changes
+		if patch.Repr.GetPatch() == "" {
```
marcelosousa, line 27: FYI: I'm not sure if this is really true. I think the same might happen with git submodules.

Anyway, this might a good approximation so I suggest we leave it as is.

For reference, it is possible to use the GitHub GraphQL API to check if a file is binary or not:

```gql
{
  repository(owner: $owner, name: $name) {
    object(expression: "BRANCH:FILEPATH") {
      ... on Blob{
        isBinary
      }
    }
  }
}
```

For example,
```gql
{
  repository(owner: "reviewpad", name: "reviewpad") {
    object(expression: "main:go.sum") {
      ... on Blob{
        isBinary
      }
    }
  }
}
```

plugins/aladino/functions/hasBinaryFile_test.go
```diff
@@ -0,0 +1,86 @@
...
+					Patch:    github.String(""),
+				},
+			},
+		},
+		"when there are two files one with no patch": {
+			wantResult: aladino.BuildBoolValue(true),
+			files: []*github.CommitFile{
+				{
+					Filename: github.String("test_bin_file"),
+					Patch:    github.String(""),
```
marcelosousa, line 61: Please provide a link to a PR that modifies a binary file so that we can confirm that it produces an empty patch.

marcelosousa (changes requested):
Thanks @zolamk ! A couple of comments from review 🙏

plugins/aladino/functions/hasBinaryFile.go
zolamk: Yeah, after a short discussion with @ferreiratiago yesterday, since this isn't a documented behavior it might be better to go with this, I'll update the PR, Thank You.

plugins/aladino/functions/hasBinaryFile_test.go
zolamk: I've updated the PR to use the GraphQL API since the patch thing ins't document behavior.

zolamk: If just checking the patch is preferable, I can revert it @marcelosousa

marcelosousa (approved):
LGTM @zolamk . Please don't forget to add a PR with the docs repository!

ferreiratiago (approved):

PR #493, opened 2022-12-02
feat: provide command error feedback as comment

### LINKED ISSUE #472: Provide reviewpad command syntax error as a new comment
Currently, when there's a syntax error on a reviewpad command, reviewpad edits the command comment (added by the user) and append the error.

Here's an example:
image

Every feedback that reviewpad needs to provide regarding the comments should be done by adding a new comment.

image

### LINKED ISSUE #473: Editing a comment with a command trigger the command
**Describe the bug**
My understanding is that commands should only run when the comment is added for the first time.

However, the command is being run every time the comment is edited.

**To Reproduce**
1. Open a project where Reviewpad is installed
2. Comment with the command `/reviewpad assign-reviewers test`
3. Reviewpad will run
4. Edit the comment added on 2.
5. Reviewpad will run again

**Expected behavior**
Reviewpad should only run when the comment is added for the first time.

**Screenshots**

zolamk:
## Description
Currently, when command errors such as syntax or invalid flag errors happen, we edit the user's comment to provide feedback, this PR modifies reviewpad so instead of editing the comment, feedback is provided as a new comment.

## Related issue

Closes #472 
Closes #473 

## Type of change

## How was this tested?
Unit & Manual Tests.

## Checklist

### DIFF
--- a/engine/exec.go
+++ b/engine/exec.go
@@ -269,12 +269,12 @@ func processCommand(env *Env, comment string) (string, error) {
 
 	err = root.Execute()
 	if err != nil {
-		comment := fmt.Sprintf("%s\n```\n🚫 error\n\n%s\n```", *env.EventData.Comment.Body, err.Error())
+		comment := fmt.Sprintf("```\n🚫 error\n\n%s```", err.Error())
 
-		if _, _, errEditComment := env.GithubClient.EditComment(env.Ctx, env.TargetEntity.Owner, env.TargetEntity.Repo, *env.EventData.Comment.ID, &github.IssueComment{
+		if _, _, errCreateComment := env.GithubClient.CreateComment(env.Ctx, env.TargetEntity.Owner, env.TargetEntity.Repo, env.TargetEntity.Number, &github.IssueComment{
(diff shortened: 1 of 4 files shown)


marcelosousa:
@zolamk here what ChatGPT has to say about your changes:

> Based on the description, the changes to the code are intended to modify the behavior of the processCommand function so that it provides feedback on command errors by creating a new comment, rather than editing the user's original comment.
> 
> In the old version of the code, the processCommand function attempts to parse the user's comment using the shellwords.Parse function, which can return an error if the comment has invalid syntax or uses an invalid flag. If this error occurs, the function constructs a new comment containing the error message and attempts to edit the user's original comment with this new comment using the GithubClient.EditComment method.
> 
> In the new version of the code, the processCommand function still attempts to parse the user's comment in the same way, but if an error occurs, it constructs a new comment containing the error message and uses the GithubClient.CreateComment method to create a new comment with this message. This means that, instead of modifying the user's original comment, the processCommand function will add a new comment providing feedback on any command errors that occur.

ferreiratiago (approved):

PR #502, opened 2022-12-05
feat: command error as comment

### LINKED ISSUE #471: Provide feedback as a comment when command fails
Currently, when a reviewpad command fails, the feedback is provided within the check run. There are several problems with this approach:
- Is not clear that the error on the check was caused by the command.
- The feedback message gets lost once the check runs again.

Here's an example:
image

The failing action/check is due to an error executing the command.

When the command fails, reviewpad should add a comment with the feedback.

zolamk:
## Description
Currently, a command might be syntactically correct but during execution of the generated program for a command an error might happen which makes it seem like it's a reviewpad error when in a reality it was the command that failed because of wrong input so this PR provides feedback to the user in the form of a GitHub comment.

## Related issue

Closes #471 

## Type of change

## How was this tested?
Unit tests and Manual tests

## Checklist


ferreiratiago:
This pull request is on hold due to low priority.

lang/aladino/env.go
```diff
@@ -51,6 +52,7 @@ type BaseEnv struct {
 	RegisterMap              RegisterMap
 	Report                   *Report
 	Target                   codehost.Target
+	EventData                *handler.EventData
```
ferreiratiago, line 55: Why do we need `EventData`? `EventPayload` should be enough for what we need here. WDYT?

engine/program.go
```diff
@@ -9,7 +9,8 @@ type Statement struct {
 }
 
 type Program struct {
-	statements []*Statement
+	IsFromCommand bool
```
ferreiratiago, line 12: `IsFromCommand` should be another property of the `env` in which the program is running. I would suggest moving this into the `BaseEnv`. WDYT?

ferreiratiago (changes requested):
Sorry for the late review. Please have a look at the comment 🙏

lang/aladino/env.go
zolamk: Yeah, makes perfect sense, thank you

engine/program.go
zolamk: This one I think better to keep it in the program since we are checking if it's a command after the `BaseEnv` is built I think, what do you think @ferreiratiago

ferreiratiago (approved):

PR #526, opened 2022-12-12
feat: merge built-in is only enabled on open and non-draft prs

### LINKED ISSUE #525: Merge action should by default be enabled only for open pull requests
As soon as we run Reviewpad on closed pull requests, it is possible that workflows conditions for the merge to continue to be valid. This can lead to unwanted behaviours where developers:
1. Try to merge multiple times the same PR;
2. Close pull requests and then add a comment (or change some PR metadata) and trigger the auto-merge workflow.

By default, the merge built-in should only be enabled if the pull request is open.

Do nothing and assume that users will always protect the merge action, e.g:
```yaml
workflos:
    - name: auto-merge-pull-requests
      always-run: true
      if:
          - $state() == "open"
      then:
          - $merge()
```

**Additional context**
A closed pull request can have the following properties:
```yml
        "mergeable": true,
        "mergeable_state": "clean",
```

This means that we can have the workflow:

```yml
    - name: auto-merge-pull-requests
      always-run: true
      if:
          - '!$isMerged() && $description() != ""'
      then:
          - $merge()
```

Consider the following sequence of actions:
1. Open a pull request with empty description
3. Close the pull request
4. Edit the description to be non-empty

It is possible that this pull request will be a state that is mergeable at which point the merge action will run and it will merge the pull request even if it was not possible to do this from the UI without reopening the pull request.

marcelosousa:
## Description

This pull request refines the $merge() built-in to only be enabled on open and non-draft pull requests.

## Related issue

Closes #525

## Type of change

## How was this tested?

Unit tests.

Functional validation with canary version of GitHub Action.

## Checklist

### DIFF
--- a/plugins/aladino/actions/merge.go
+++ b/plugins/aladino/actions/merge.go
@@ -6,6 +6,7 @@ package plugins_aladino_actions
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
 	"github.com/reviewpad/reviewpad/v3/handler"
@@ -22,6 +23,11 @@ func Merge() *aladino.BuiltInAction {
 
 func mergeCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
+
+	if t.PullRequest.GetState() != "open" || t.PullRequest.GetDraft() {
+		log.Println("merge: skipping action because pull request is not open or is a draft")
+		return nil
+	}
 
 	mergeMethod, err := parseMergeMethod(args)
 	if err != nil {
--- a/plugins/aladino/actions/merge_test.go
+++ b/plugins/aladino/actions/merge_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/google/go-github/v48/github"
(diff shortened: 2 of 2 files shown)


plugins/aladino/actions/merge.go
```diff
@@ -23,6 +24,11 @@ func Merge() *aladino.BuiltInAction {
 func mergeCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
 
+	if t.PullRequest.GetState() != "open" || t.PullRequest.GetDraft() {
+		log.Println("merge: skipping action because pull request is not open or is a draft")
+		return nil
+	}
```
ferreiratiago, line 30: I would suggest splitting this condition into two.
```suggestion
	if t.PullRequest.GetDraft() {
		log.Println("merge: skipping action because pull request is on draft")
		return nil
	}

	if t.PullRequest.GetState() != "open" {
		log.Println("merge: skipping action because pull request is not open")
		return nil
	}
```

WDYT?

ferreiratiago (approved):
Minor suggestion but not blocking.

plugins/aladino/actions/merge.go
marcelosousa: Lets leave this as is.

Issue #538, opened 2022-12-14
Add built-in filesPath


ferreiratiago:
Add the new built-in `filesPath` that follows the signature `filesPath(): string[]` and returns the file paths of all changed files in the patch.

PR #542, opened 2022-12-15
feat: join built-in function

zoldater:
## Description

The pull request introduces the `$join()` built-in function that returns a string value concatenating array elements with separator.

## Motivation

`sprintf` function replaces existing specifiers with the values from `arguments_list `. For this reason passing `$group()`, `$team()`, `$assignees()` or any array of arbitrary size as `arguments_list` could be correct.

Proposed change makes it possible to join arrays into string to replace only one format specifier.

## Example

- `$sprintf("These changes require approval by team 'a': %s", [$join($group("a"))])`

## Type of change

## How was this tested?

Unit tests.

## Checklist

### DIFF
--- a/engine/transform.go
+++ b/engine/transform.go
@@ -73,6 +73,21 @@ func addDefaultHaveAllChecksRunCompleted(str string) string {
 	return r.ReplaceAllString(str, `$$haveAllChecksRunCompleted([], "")`)
 }
 
+func addDefaultJoinSeparator(str string) string {
+	r := regexp.MustCompile(`\$join\((\[.*\]|([^,]*))(,\s*"([^"]*)")?\)`)
+	match := r.FindStringSubmatch(str)
+	if len(match) == 0 {
+		return str
+	}
+
+	toJoin := match[1]
+	separator := match[4]
+	if separator == "" {
+		separator = " "
+	}
+	return strings.Replace(str, match[0], "$join("+addDefaultJoinSeparator(toJoin)+", \""+separator+"\")", 1)
+}
+
 func transformAladinoExpression(str string) string {
 	transformedActionStr := str
 
@@ -85,6 +100,7 @@ func transformAladinoExpression(str string) string {
 		addEmptyCloseComment,
 		addDefaultCloseReason,
 		addDefaultHaveAllChecksRunCompleted,
+		addDefaultJoinSeparator,
 	}
 
 	for i := range transformations {
--- a/engine/transform_internal_test.go
+++ b/engine/transform_internal_test.go
@@ -115,6 +115,50 @@ func TestTransformAladinoExpression(t *testing.T) {
 			arg:     `$haveAllChecksRunCompleted() && true && $haveAllChecksRunCompleted(["build"]) && $haveAllChecksRunCompleted(["build", "run"], "completed") && $haveAllChecksRunCompleted([]) && $haveAllChecksRunCompleted([], "")`,
 			wantVal: `$haveAllChecksRunCompleted([], "") && true && $haveAllChecksRunCompleted(["build"], "") && $haveAllChecksRunCompleted(["build", "run"], "completed") && $haveAllChecksRunCompleted([], "") && $haveAllChecksRunCompleted([], "")`,
 		},
+		"join empty array with empty separator": {
+			arg:     `$join([])`,
+			wantVal: `$join([], " ")`,
+		},
+		"join empty array with non-empty separator": {
+			arg:     `$join([], ", ")`,
+			wantVal: `$join([], ", ")`,
+		},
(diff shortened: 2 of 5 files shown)


engine/transform_internal_test.go
```diff
@@ -115,6 +115,46 @@ func TestTransformAladinoExpression(t *testing.T) {
...
+			wantVal: `$$join($team("a"), ", ")`,
+		},
+		"join assignees with empty separator": {
+			arg:     `$join($assignees())`,
+			wantVal: `$join($assignees(), " ")`,
+		},
+		"join assignees with non-empty separator": {
+			arg:     `$join($assignees(), ", ")`,
+			wantVal: `$join($assignees(), ", ")`,
+		},
```
marcelosousa, line 157: The transform has a limitation that it doesn't work for nested $join statements. For example:

```
        "nested join with empty separators ": {
			arg:     `$join([$join(["a","b"]), "c"])`,
			wantVal: `$join([$join(["a","b"], " "), "c"], " ")`,
		},
```

marcelosousa (commented):
Many thanks for this PR 🙌  I've left a comment related to the transform of the join built-in as it is not working for nested calls. I believe this is a broader issue of the way we are doing these transformations and should be addressed in a different issue. @ferreiratiago should I open an issue for that so that we can unblock this PR?

For the rest, it's looking good - thanks!

marcelosousa:
Test coverage is good!

engine/transform_internal_test.go
zoldater: Thanks @marcelosousa !
I've added the test and fixed issue with recursion.

zoldater:
Hi @ferreiratiago and @marcelosousa !
I fixed the issue with nested join calls and added proposed test to `engine/transform_internal_test.go`. So it should not be necessary to open new issue for that.

engine/transform.go
```diff
@@ -73,6 +73,21 @@ func addDefaultHaveAllChecksRunCompleted(str string) string {
 	return r.ReplaceAllString(str, `$$haveAllChecksRunCompleted([], "")`)
 }
 
+func addDefaultJoinSeparator(str string) string {
+	r := regexp.MustCompile(`\$join\(((\[.*\])|(\$(team|group)\("[^"]*"\))|(\$assignees\(\)))(,\s*"([^"]*)")?\)`)
```
ferreiratiago, line 77: What's the reasoning behind `team`, `group`, and `assignees`? I may be missing something 😅

Isn't a simpler regex enough?
```suggestion
	r := regexp.MustCompile(`\$join\((\[.*\]|([^,]*))(,\s*"([^"]*)")?\)`)
```

If we have the literal values `team`, `group`, and `assignees`, we need to keep updating this transform every time there's a new built-in that returns a []string. For instance, [$filePaths()](https://github.com/reviewpad/reviewpad/issues/538).

ferreiratiago (changes requested):

plugins/aladino/functions/join.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(elementsArg.Vals) == 0 {
+		return aladino.BuildStringValue(""), nil
+	}
+	var clearVals []string
+	separatorArg := args[1].(*aladino.StringValue)
+	for _, val := range elementsArg.Vals {
+		switch v := val.(type) {
+		case *aladino.StringValue:
+			clearVals = append(clearVals, v.Val)
+		}
```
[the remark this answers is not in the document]
zoldater: Done.

engine/transform.go
zoldater: Thanks for pointing this out! Simplified the regex.

plugins/aladino/functions/join.go
```diff
@@ -0,0 +1,40 @@
...
+		return aladino.BuildStringValue(""), nil
+	}
+	var clearVals []string
+	separatorArg := args[1].(*aladino.StringValue)
+	for _, val := range elementsArg.Vals {
+		switch v := val.(type) {
+		case *aladino.StringValue:
+			clearVals = append(clearVals, v.Val)
+		default:
+			return nil, fmt.Errorf("join: invalid element of kind %v", v.Kind())
```
ferreiratiago, line 35: We are very close. Just missing the unit test for this case.

ferreiratiago (changes requested):
One unit test missing. Other than that it's perfect 🙌

plugins/aladino/functions/join.go
zoldater: Added `TestJoin_BadType`

ferreiratiago (approved):
Perfect. Thanks for your contribution @zoldater 🙏

PR #553, opened 2022-12-16
feat: update collector with options

ferreiratiago:
## Description

This pull request introduces the parameter options to collector and the new method collect error.

This is to allow the collect to be created with the minium values.

## Related to

Closes https://github.com/reviewpad/reviewpad/issues/554

## Type of change

## How was this tested?

Run `task check -f`

## Checklist

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -11,42 +11,67 @@ import (
 
 type Collector interface {
 	Collect(eventName string, properties map[string]interface{}) error
+	CollectError(err error) error
 }
 
 type collector struct {
-	Client    mixpanel.Mixpanel
-	Id        string
-	Token     string
+	Client mixpanel.Mixpanel
+	// Unique identifier that is connected to every event.
+	// It is used to identify the user.
+	DistinctId string
+	// Token to authenticate the requests and identify the project.
+	Token string
+	// Runner environment where the event is being collected.
+	// (i.e. github app, github action, playground )
 	RunnerEnv string
-	// Allows to identify a unique source of events
+	// Unique identifier for the events' source.
 	RunnerId string
-	// Allows to identify the correct order of events
+	// Order of the event in the events' source.
 	Order int
-	// Allows to identify the type of the event
+	// Type of the event.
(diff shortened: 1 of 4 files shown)


collector/collector.go
```diff
@@ -11,60 +11,101 @@ import (
...
-	// Allows to identify the url
+	// Optional properties that are added to every event.
+	Optional *OptionalProperties
+}
+
+type OptionalProperties struct {
+	// The repository url where the events are being collected.
 	Url string
+	// Delivery ID of the event.
+	DeliveryId string
```
marcelosousa, line 42: What's the difference between the `DeliveryId` and the `DistinctId`?

collector/collector.go
```diff
@@ -11,60 +11,101 @@ import (
...
-	Id        string
-	Token     string
+	Client mixpanel.Mixpanel
+	// Unique identifier that is connected to every event.
+	// It is used to identify the user.
+	DistinctId string
+	// Token to authenticate the requests and identify the project.
+	Token string
+	// Runner environment where the event is being collected.
+	// (i.e. github app, github action, playground )
```
marcelosousa, line 25:
```suggestion
	// (i.e. github app, github action, playground)
```

marcelosousa (approved):
I made a couple of comments in the diff. It's not easy to fully understand these changes and how to apply them in the consumers. It would be good to see the PR in the action repository that will use this new consumer. 
Looking at https://github.com/reviewpad/reviewpad/issues/554, it could be nice for future documentation purpose to list a couple of examples of where the `url` is not required; I thought this was always the case so that we can keep track of the entire journey. 
I think I'm missing something here but IMO it doesn't deserve blocking this PR so I'm approving the PR!

collector/collector.go
ferreiratiago: `DistinctId` it is used to identify the user.
`DeliveryId` it is used to identify the event within the runner. For instance, in both github action and app it corresponds to the "X-GitHub-Delivery".

PR #558, opened 2022-12-20
feat: context built-in function

zolamk:
## Description
Adds a new built-in function `$context` that returns a `JSON` serialized string presentation of the target(pull request / issue).

## Type of change

## How was this tested?
Unit tests and manual tests

## Checklist

- [x] Ask: this pull request requires a code review before merge -->

### DIFF
--- a/plugins/aladino/functions/context.go
+++ b/plugins/aladino/functions/context.go
@@ -9,19 +9,19 @@ import (
 	"github.com/reviewpad/reviewpad/v3/lang/aladino"
 )
 
-func Content() *aladino.BuiltInFunction {
+func Context() *aladino.BuiltInFunction {
 	return &aladino.BuiltInFunction{
 		Type:           aladino.BuildFunctionType([]aladino.Type{}, aladino.BuildStringType()),
-		Code:           contentCode,
+		Code:           contextCode,
 		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest, handler.Issue},
 	}
 }
 
-func contentCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
-	content, err := e.GetTarget().JSON()
+func contextCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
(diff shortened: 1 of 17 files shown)


ferreiratiago (approved):

plugins/aladino/functions/context_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.Nil(t, err)
+
+	tests := map[string]struct {
+		targetEntity *handler.TargetEntity
+		wantErr      error
+		wantRes      aladino.Value
+	}{
+		"when pull request": {
+			targetEntity: aladino.DefaultMockTargetEntity,
+			wantRes:      aladino.BuildStringValue(string(mockPRJSON)),
```
marcelosousa, line 33: IMO, these unit tests are not enough functionally.

As a dev, I don't know exactly what is the string format that is being matched.

Please add a couple of tests that check against actual mocked strings so that we can guarantee that all understand the string format of that is returned by $context().

This is important because if we want to some regex over it, we need to understand it's format.

marcelosousa (changes requested):
Thanks for this PR - please check my comment in the context_text.go file.

plugins/aladino/functions/context.go
```diff
@@ -0,0 +1,27 @@
...
+
+func Context() *aladino.BuiltInFunction {
+	return &aladino.BuiltInFunction{
+		Type:           aladino.BuildFunctionType([]aladino.Type{}, aladino.BuildStringType()),
+		Code:           contextCode,
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest, handler.Issue},
+	}
+}
+
+func contextCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
```
marcelosousa, line 20: I think we should use the blank identifier here to state that the variable or argument is being intentionally ignored.

```suggestion
func contextCode(e aladino.Env, _ []aladino.Value) (aladino.Value, error) {
```

marcelosousa (approved):

PR #561, opened 2022-12-21
feat: toJSON built-in function

zolamk:
## Description
Adds a new `$toJSON` built in function that takes a string and returns a json type.

## Type of change

## How was this tested?
Unit tests

## Checklist

- [x] Ask: this pull request requires a code review before merge -->


lang/aladino/value.go
```diff
@@ -186,3 +187,25 @@ func (thisVal *FunctionValue) Equals(other Value) bool {
...
+func (jVal *JSONValue) Kind() string {
+	return JSON_VALUE
+}
+
+func (jVal *JSONValue) HasKindOf(ty string) bool {
+	return jVal.Kind() == ty
+}
+
+func (jVal *JSONValue) Equals(other Value) bool {
+	// TODO: fix equals
```
ferreiratiago, line 209: Can you please open an issue to address this later on and add it as a comment here?

ferreiratiago (approved):

lang/aladino/type.go
```diff
@@ -54,6 +57,10 @@ func BuildArrayType(elemsTypes []Type) *ArrayType {
 	return &ArrayType{elemsTypes}
 }
 
+func BuildJSONType() *JSONType {
+	return &JSONType{}
+}
```
marcelosousa, line 62: Turn into an one liner as in L44

lang/aladino/value.go
```diff
@@ -186,3 +189,28 @@ func (thisVal *FunctionValue) Equals(other Value) bool {
 func (fVal *FunctionValue) HasKindOf(ty string) bool {
 	return fVal.Kind() == ty
 }
+
+// JSONValue represents a json value
+type JSONValue struct {
+	Val interface{}
```
marcelosousa, line 195: Why not use `json.RawMessage` instead of `interface{}` here? As it stands, `JSONValue` can represent any Go value which is not really what is intended. Maybe I'm missing something here :-)

lang/aladino/value_test.go
```diff
@@ -334,3 +334,74 @@ func TestFunctionValueEquals_WhenFalse(t *testing.T) {
...
+
+func TestJSONValueEquals(t *testing.T) {
+	tests := map[string]struct {
+		x      *aladino.JSONValue
+		y      *aladino.JSONValue
+		equals bool
+	}{
+		"when types are the same but values differ": {
+			x:      aladino.BuildJSONValue(1),
+			y:      aladino.BuildJSONValue(10),
```
marcelosousa, line 346: I'm confused as why you would want to do this.

marcelosousa (changes requested):
Please check the comment about the choice of `Val` in the `JSONValue`

marcelosousa:
@ferreiratiago @zolamk I have a question to ensure that we are aligned regarding JSON values in Aladino. From what I can understand, the idea is for JSON values to be a kind of internal aladino value for now correct? What is mean concretely with this question is that the only way to build a JSONValue would be use the built-in `$toJSON(...)` with a string.

plugins/aladino/functions/toJSON.go
```diff
@@ -0,0 +1,30 @@
...
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest, handler.Issue},
+	}
+}
+
+func toJSONCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
+	jsonString := args[0].(*aladino.StringValue).Val
+
+	val, err := oj.ParseString(jsonString)
+	if err != nil {
+		return nil, err
```
marcelosousa, line 26: This is an issue for the future but we shouldn't be simply yielding errors that come from 3rd party libraries. We should be parsing those errors and converting them to our format so that users can have consistent errors messages.  (cc @ferreiratiago @zolamk @shay2025)
@ferreiratiago could we open an issue for that before merging this PR?

marcelosousa: Issue created

lang/aladino/value.go
zolamk: The library we are using https://github.com/ohler55/ojg for parsing expressions and evaluating them uses the `interface{}` type, but I think we refactor it to work with `json.RawMessage` later on maybe, what do you think?

lang/aladino/value_test.go
zolamk: Just a test case but integers are valid JSON as well i think.

lang/aladino/value.go
marcelosousa: Alright, makes sense!

marcelosousa (approved):

PR #570, opened 2022-12-22
feat: add logging info for overrides

### LINKED ISSUE #517: Add logging information for overrides
In connection with #306 and #515, we need to add logging information to notify users that labels, groups, rules, workflows or pipelines were overridden.

shay2025:
## Description

This pull request introduces the logging for overridden reviewpad properties by the extend property.
This logging says which properties were overridden and the property type.

## Related issue

Closes #517

## Type of change

## How was this tested?

Ran manual tests with `reviewpad-cli`.

## Checklist

### DIFF
--- a/engine/loader.go
+++ b/engine/loader.go
@@ -9,6 +9,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 
 	gh "github.com/reviewpad/reviewpad/v3/codehost/github"
@@ -20,6 +21,11 @@ import (
 type LoadEnv struct {
 	Visited map[string]bool
 	Stack   map[string]bool
+}
+
+type OverridenElem struct {
+	PropertyType            string
+	overridenIntroducedFile string
 }
 
 func hash(data []byte) string {
@@ -51,9 +57,17 @@ func Load(ctx context.Context, githubClient *gh.GithubClient, data []byte) (*Rev
 		return nil, err
 	}
 
-	file, err = processExtends(ctx, githubClient, file, env)
+	overridenElems := make(map[string](map[string]*OverridenElem))
+	file, err = processExtends(ctx, githubClient, file, env, overridenElems)
 	if err != nil {
 		return nil, err
+	}
+
+	for _, overridenElem := range overridenElems {
+		for elemName, elemInfo := range overridenElem {
+			log.Println(overridenElem)
+			log.Printf("[WARN] the %s %s has been overriden after extending the configuration in %s", elemInfo.PropertyType, elemName, elemInfo.overridenIntroducedFile)
+		}
 	}
 
 	file, err = normalize(file, inlineRulesNormalizer())
@@ -256,7 +270,7 @@ func loadExtension(ctx context.Context, githubClient *gh.GithubClient, extension
 // processExtends inlines files into the current reviewpad file
 // precedence: current file > extends file
 // Post-condition: ReviewpadFile without extends statements
-func processExtends(ctx context.Context, githubClient *gh.GithubClient, file *ReviewpadFile, env *LoadEnv) (*ReviewpadFile, error) {
+func processExtends(ctx context.Context, githubClient *gh.GithubClient, file *ReviewpadFile, env *LoadEnv, overridenElems map[string](map[string]*OverridenElem)) (*ReviewpadFile, error) {
 	extendedFile := &ReviewpadFile{}
 	for _, extensionUri := range file.Extends {
 		eFile, eHash, err := loadExtension(ctx, githubClient, extensionUri)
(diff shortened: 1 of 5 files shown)


ferreiratiago (changes requested):
Unfortunately, this solution does not work properly. The proposed solution reimplements the "extends" algorithm, which is ok until the algorithm changes. Also, we need to be able to provide feedback for all properties of all the files. The current implementation does not take into account the last file, e.g. for A -> B -> C, the current solution will log the extends B -> C but not A -> B, where A is the `reviewpad.yml` at the root of the repository.

The ideal solution would be, for each relation A -> B and B -> C, when there's a property in common, to share which extension had the priority. For instance,

For the extensions `A -> B -> C`. Let's assume they all have the same rule `X`.

The outputted logs should be:
```bash
the rule X has been overridden by configuration B
the rule X has been overridden by configuration A
```

Also, please, use the `logger` (available on `env`) and a grammar checker (there are grammar errors such as `overridens`).

engine/loader.go
```diff
@@ -22,13 +23,60 @@ type LoadEnv struct {
...
+
+	for i := len(extensionFilesStack) - 1; i >= 1; i-- {
+		extendedUri := extensionFilesStack[i].uri
+		extendedFile := extensionFilesStack[i].file
+
+		extenderUri := extensionFilesStack[i-1].uri
+		if extenderUri == "" {
+			extenderUri = "the root configuration"
+		}
+		extenderFile := extensionFilesStack[i-1].file
```
ferreiratiago, line 44: This logic is really hard to understand. It relies in the order of the elements and that's not clear.

engine/loader.go
```diff
@@ -254,21 +310,26 @@ func loadExtension(ctx context.Context, githubClient *gh.GithubClient, extension
 // processExtends inlines files into the current reviewpad file
 // precedence: current file > extends file
 // Post-condition: ReviewpadFile without extends statements
-func processExtends(ctx context.Context, githubClient *gh.GithubClient, file *ReviewpadFile, env *LoadEnv) (*ReviewpadFile, error) {
+func processExtends(ctx context.Context, logger *logrus.Entry, githubClient *gh.GithubClient, file *ReviewpadFile, env *LoadEnv, extendedFiles *[]ReviewpadFileWithUri) (*ReviewpadFile, error) {
```
ferreiratiago, line 313: I don't think logger is being used.

engine/loader.go
```diff
@@ -51,11 +99,19 @@ func Load(ctx context.Context, githubClient *gh.GithubClient, data []byte) (*Rev
 		return nil, err
 	}
 
-	file, err = processExtends(ctx, githubClient, file, env)
+	extensionFilesStack := []ReviewpadFileWithUri{
+		{
+			uri:  "",
+			file: file,
+		},
+	}
```
ferreiratiago, line 107: Looking into the naming I'm expecting the list to be of the extended files. But it starts by adding the file at the root, which in not extended. It can be confusing.

engine/loader.go
```diff
@@ -22,13 +23,60 @@ type LoadEnv struct {
...
+
+		extenderUri := extensionFilesStack[i-1].uri
+		if extenderUri == "" {
+			extenderUri = "the root configuration"
+		}
+		extenderFile := extensionFilesStack[i-1].file
+
+		overridenLabels := getOverridenLabels(extenderFile, extendedFile)
+		for _, overridenLabel := range overridenLabels {
+			log.Warnf("the label %s in %s has been overriden by the label in %s", overridenLabel, extendedUri, extenderUri)
```
ferreiratiago, line 48: Typo: `overriden` should be `overridden`. Please use a spell checker 🙏

engine/loader.go
```diff
@@ -22,13 +23,60 @@ type LoadEnv struct {
 	Stack   map[string]bool
 }
 
+type ReviewpadFileWithUri struct {
+	uri  string
+	file *ReviewpadFile
+}
+
+func logOverrides(log *logrus.Entry, extensionFilesStack []ReviewpadFileWithUri) {
```
ferreiratiago, line 31: I think we can simplify this function. Can you please an approach similar to this:

```go
func logExtendedProperties(logger *logrus.Entry, fileA *ReviewpadFile, fileAUri string, fileB *ReviewpadFile, fileBUri string, resultFile *ReviewpadFile) {
	// Extended labels
	for _, label := range resultFile.Labels {
		if _, ok := fileA.Labels[label.Name]; ok {
			if _, ok := fileB.Labels[label.Name]; ok {
				if fileA.Labels[label.Name].equals(label) {
					logger.Warnf("label %s has been overridden by %s", label.Name, fileAUri)
				} else {
					logger.Warnf("label %s has been overridden by %s", label.Name, fileBUri)
				}
			}
		}
	}
	// Extended groups
	// Extended rules
	// Extended workflows
	// Extended pipelines
}
```

Follow this suggestion we can call this function inside `processExtends` after `extendedFile.extend(extensionFile)`.

```go
func processExtends(ctx context.Context, logger *logrus.Entry, githubClient *gh.GithubClient, file *ReviewpadFile, fileUri string, env *LoadEnv) (*ReviewpadFile, error) {
	extendedFile := &ReviewpadFile{}
	for _, extensionUri := range file.Extends {
		// ...

		// remove from the stack
		delete(env.Stack, eHash)

		extendedFile.extend(extensionFile)

		// log here too
		logExtendedProperties(logger, file, fileUri, extensionFile, extensionUri, extendedFile)
	}

	extendedFile.extend(file)

	// log here too

	// reset all extends
	extendedFile.Extends = []string{}

	return extendedFile, nil
}
```

ferreiratiago (changes requested):

ferreiratiago (approved):

PR #593, opened 2022-12-27
fix: join separator default function

### LINKED ISSUE #583: Error on using sprintf with ArrayType arguments
**Describe the bug**
The following aladino expression: 
```
$warn($sprintf("Request a 1:1 with security team: %s", [$join(["marcelosousa", "ferreiratiago"], " - ")]))
``` 

Is not being correctly handled by interpreter because we get the error: `type inference failed: mismatch in arg types on sprintf`.

I already checked that the AST seems to be correct so the investigation needs to focus in the type inference.

I think the root cause is that `sprintf` is expecting an `ArrayOfType(String)` while the type of `["marcelosousa", "ferreiratiago"]` is an `ArrayType(String)`. 

This is likely the cause because the following expression works:

```
$warn($sprintf("Request a 1:1 with security team: %s", [$group("team", " - ")]))
``` 

Because `$group()` returns an `ArrayOfType(String)`.

zolamk:
## Description
Fixes the `$join` built-in functions default separator transformer.

The problem was the `addDefaultJoinSeparator` function, in order to support nested joins used a regular expression that caused it to add the default separator in the wrong place for certain conditions. Unfortunately, this fix has the side effect of causing the default function not to provide defaults for nested joins.

## Related issue

Closes #583 

## Type of change

## How was this tested?
Unit tests and manual tests

## Checklist

### DIFF
--- a/engine/transform.go
+++ b/engine/transform.go
@@ -74,18 +74,10 @@ func addDefaultHaveAllChecksRunCompleted(str string) string {
 }
 
 func addDefaultJoinSeparator(str string) string {
-	r := regexp.MustCompile(`\$join\((\[.*\]|([^,]*))(,\s*"([^"]*)")?\)`)
-	match := r.FindStringSubmatch(str)
-	if len(match) == 0 {
-		return str
-	}
-
-	toJoin := match[1]
-	separator := match[4]
-	if separator == "" {
-		separator = " "
-	}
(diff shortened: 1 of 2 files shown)


ferreiratiago (approved):

ferreiratiago:
> The problem was the addDefaultJoinSeparator function, in order to support nested joins used a regular expression that caused it to add the default separator in the wrong place for certain conditions. Unfortunately, this fix has the side effect of causing the default function not to provide defaults for nested joins.

Can we update the documentation with this limitation? And also share some examples of uses of nested joins that won't work.

PR #613, opened 2023-01-13
feat: get mixpanel user

shay2025:
## Description

This pull request introduces a method that queries for user profile data given a specific distinct id. This pull request was done to support the feature of collecting Reviewpad GitHub Action installation events.

## Type of change

## How was this tested?

Manual tests and ran `task check` cmd.

## Checklist


mixpanel/query.go
```diff
@@ -0,0 +1,57 @@
...
+	Page      int           `json:"page"`
+	PageSize  int           `json:"page_size"`
+	SessionId string        `json:"session_id"`
+	Status    string        `json:"status"`
+	Total     int           `json:"total"`
+	Results   []interface{} `json:"results"`
+}
+
+func GetMixpanelUser(distinctId, mixpanelAPISecret, mixpanelPassword string) (*QueryProfilesResponse, error) {
+	url := "https://eu.mixpanel.com/api/2.0/engage?project_id=2729960"
```
ferreiratiago, line 26: Let's add the project id as an env variable. Otherwise, we can't use it for other environments.

ferreiratiago (changes requested):
Please move this code to the action since this is only used there.

mixpanel/query.go
shay2025: @ferreiratiago The pull request has been moved.

PR #656, opened 2023-01-31
fix: ignore reviewpad check runs by default

### LINKED ISSUE #655: $haveAllChecksRunCompleted built-in does not ignore reviewpad check
**Describe the bug**
Currently, the built-in `haveAllChecksRunCompleted` does not ignore the reviewpad check. This causes the built-in to always return `false` since the Reviewpad check is always running when the built-in is triggered.
 
**To Reproduce**
1. Define a Reviewpad configuration that uses the built-in `haveAllChecksRunCompleted`
2. Open a pull request to trigger Reviewpad
3. You will see that the workflow using the built-in `haveAllChecksRunCompleted` will never be triggered regardless of the check run status.

**Expected behavior**
The built-in `haveAllChecksRunCompleted` should ignore the Reviewpad check in order to work properly.

ferreiratiago:
## Description

This pull request introduces the fix to the built-in `haveAllChecksRunCompleted` by ignoring the Reviewpad check by default.

## Related issue

Closes #655

## Type of change

## How was this tested?

Manuel tests

## Checklist

### DIFF
--- a/plugins/aladino/functions/haveAllChecksRunCompleted_test.go
+++ b/plugins/aladino/functions/haveAllChecksRunCompleted_test.go
@@ -389,6 +389,49 @@ func TestHaveAllChecksRunCompleted(t *testing.T) {
 			},
 			wantResult: aladino.BuildBoolValue(true),
 		},
+		"when reviewpad is one of the checks": {
+			checkRunsToIgnore: aladino.BuildArrayValue([]aladino.Value{}),
+			conclusion:        aladino.BuildStringValue(""),
+			mockBackendOptions: []mock.MockBackendOption{
+				mock.WithRequestMatch(
+					mock.GetReposCommitsCheckRunsByOwnerByRepoByRef,
+					github.ListCheckRunsResults{
+						Total: github.Int(2),
+						CheckRuns: []*github.CheckRun{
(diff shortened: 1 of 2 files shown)


zolamk (approved):

plugins/aladino/functions/haveAllChecksRunCompleted_test.go
```diff
@@ -389,6 +389,49 @@ func TestHaveAllChecksRunCompleted(t *testing.T) {
 			},
 			wantResult: aladino.BuildBoolValue(true),
 		},
+		"when reviewpad is one of the checks": {
```
shay2025, line 392: Shouldn't we have two test cases where reviewpad is one of the checks and:
- the check runs have all been completed;
- not all check runs have been completed.

shay2025 (approved):
Just a thought

plugins/aladino/functions/haveAllChecksRunCompleted_test.go
ferreiratiago: I see. There's no need for that. The important thing here is for the Reviewpad check to be ignored regardless of the other checks' status.

PR #657, opened 2023-01-31
feat: add account type and visibility to target entity

### LINKED ISSUE #651: Update collector data
For each event being collected we need to know:
- [ ] Account type - organization / user
- [ ] Visibility - public / private

zolamk:
## Description
Adds `account type` and `repository visibility` to target entity so that we can use it when collecting events to mixpanel.

## Related issue

Closes #651 

## Type of change

## How was this tested?
Unit tests

## Checklist

### DIFF
--- a/handler/processors.go
+++ b/handler/processors.go
@@ -32,10 +32,12 @@ func (entityType TargetEntityKind) String() string {
 }
 
 type TargetEntity struct {
-	Kind   TargetEntityKind
-	Number int
-	Owner  string
-	Repo   string
+	Kind        TargetEntityKind
+	Number      int
+	Owner       string
+	Repo        string
+	AccountType string
+	Visibility  string
 }
 
 type EventDetails struct {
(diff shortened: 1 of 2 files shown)


handler/processors.go
```diff
@@ -88,10 +90,12 @@ func processCronEvent(log *logrus.Entry, token string, e *ActionEvent) ([]*Targe
...
-			Kind:   kind,
-			Number: *issue.Number,
-			Owner:  owner,
-			Repo:   repo,
+			Kind:        kind,
+			Number:      *issue.Number,
+			Owner:       owner,
+			Repo:        repo,
+			AccountType: issue.GetRepository().GetOwner().GetType(),
+			Visibility:  issue.GetRepository().GetVisibility(),
```
ferreiratiago, line 98: I'm not sure if `Visibility` is always present. Can you please make sure? Here's how we do it in the [action](https://github.com/reviewpad/action/blob/main/agent/collector.go#L18).

ferreiratiago (changes requested):

shay2025 (changes requested):
It might make sense to lowercase the account type in order to stay consistent with the other events that are collected and display the account type, e.g., https://github.com/reviewpad/action/blob/main/agent/collector.go#L46 and on the RAAS.

ferreiratiago (approved):

zolamk:
> Member

I think we can lower it on the raas and action side

zolamk:
> It might make sense to lowercase the account type in order to stay consistent with the other events that are collected and display the account type, e.g., https://github.com/reviewpad/action/blob/main/agent/collector.go#L46 and on the RAAS.

I think we can lower it on the raas and action side

shay2025:
> > It might make sense to lowercase the account type in order to stay consistent with the other events that are collected and display the account type, e.g., https://github.com/reviewpad/action/blob/main/agent/collector.go#L46 and on the RAAS.
> 
> I think we can lower it on the raas and action side

On the RAAS and action side, it is lowered already 🙂

PR #660, opened 2023-02-01
feat: extend config unauthorized err handling

### LINKED ISSUE #659: Extend configuration from an unauthorized repository
Currently, when a Reviewpad configuration extends a configuration from a repository from which there are no permissions to fetch the load will fail.

For example, if configuration A extends from configuration B and Reviewpad is installed on repository A but not on B, the load of the Reviewpda configuration will fail with the message similar to `error loading reviewpad file: GET https://api.github.com/repos/ferreiratiago/.github/contents/reviewpad-config?ref=main: 404 Not Found [`

We should be able to detect that the error is an authorization error and fail Reviewpad with a proper message.

Something similar to:
`We were unable to load the Reviewpad configuration due to an authorization error. If you are using extends please make sure you have Reviewpad installed in all extended repositories`

shay2025:
## Description

This pull request implements a feature to detect when a "404 Not Found" error is due to reviewpad not having access to the `reviewpad.yml` files that come from the `extends` property. The code checks the error message from the GitHub API after trying to download a file from a repository. If the error message is "Not Found" or "Repository access blocked", the error is likely due to reviewpad not being installed on a private repository. If it's not due to this reason, the error may be due to other factors.

It's worth noting that "404 Not Found" is a general HTTP error indicating that the requested resource could not be found on the server. On the other hand, "Repository access blocked" is a specific error message generated by GitHub indicating the user doesn't have sufficient permissions to access the requested repository.

GitHub uses "404 Not Found" instead of "Repository access blocked" when a user tries to download a file from a private repository they don't have access to because it fits the general case of a resource not being found on the server, regardless of the reason why it's unavailable. This allows for consistency in error messaging and avoids the need for different error messages for every possible reason a resource may be unavailable.

## Related issue

Closes #659 

## Type of change

## How was this tested?

Manual tests (described in https://www.notion.so/reviewpad/Extend-configuration-from-an-unauthorized-repository-85fbe0b78ace4ebe8062b283e9ef7d76) and ran the `task check` cmd.

## Checklist

### DIFF
--- a/engine/loader.go
+++ b/engine/loader.go
@@ -354,8 +354,8 @@ func loadExtension(ctx context.Context, githubClient *gh.GithubClient, extension
 	content, err := githubClient.DownloadContents(ctx, filePath, branch)
 	if err != nil {
 		if responseErr, ok := err.(*github.ErrorResponse); ok && responseErr.Response.StatusCode == 404 {
-			if responseErr.Message == "Repository access blocked" {
-				return nil, "", fmt.Errorf("We were unable to load the Reviewpad configuration due to an authorization error. If you are using extends please make sure you have Reviewpad installed in all extended repositories.")
+			if responseErr.Message == "Not Found" {
+				return nil, "", fmt.Errorf("We encountered an error while loading the Reviewpad configuration. This may be due to reasons such as incorrect file path, invalid URL, or unauthenticated access. Please verify that you have installed Reviewpad in all extended repositories and that the file name and path are correct. If you still encounter this error, reach out to us at #help on https://reviewpad.com/discord for additional support.")
 			} else {
 				return nil, "", err
 			}
--- a/engine/loader.go
+++ b/engine/loader.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/google/go-github/v49/github"
 	gh "github.com/reviewpad/reviewpad/v3/codehost/github"
 	"github.com/reviewpad/reviewpad/v3/handler"
 	"github.com/reviewpad/reviewpad/v3/utils"
@@ -352,7 +353,13 @@ func loadExtension(ctx context.Context, githubClient *gh.GithubClient, extension
 
 	content, err := githubClient.DownloadContents(ctx, filePath, branch)
 	if err != nil {
-		return nil, "", err
+		if responseErr, ok := err.(*github.ErrorResponse); ok && responseErr.Response.StatusCode == 404 {
+			if responseErr.Message == "Repository access blocked" {
+				return nil, "", fmt.Errorf("We were unable to load the Reviewpad configuration due to an authorization error. If you are using extends please make sure you have Reviewpad installed in all extended repositories.")
+			} else {
+				return nil, "", err
+			}
+		}
 	}
 
 	file, err := parse(content)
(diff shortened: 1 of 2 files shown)


engine/loader.go
```diff
@@ -352,7 +353,16 @@ func loadExtension(ctx context.Context, githubClient *gh.GithubClient, extension
...
+		if responseErr, ok := err.(*github.ErrorResponse); ok && responseErr.Response.StatusCode == 404 {
+			switch responseErr.Message {
+			case "Not Found":
+				return nil, "", fmt.Errorf("We encountered an error while loading the Reviewpad configuration. This may be due to reasons such as incorrect file path, invalid URL, or unauthenticated access. Please verify that you have installed Reviewpad in all extended repositories and that the file name and path are correct. If you still encounter this error, reach out to us at #help on https://reviewpad.com/discord for additional support.")
+			case "Repository access blocked":
+				return nil, "", fmt.Errorf("We encountered an authorization error while trying to load the Reviewpad configuration. Please ensure that Reviewpad is installed in all extended repositories if you are using the 'extends' feature. If the issue persists, kindly reach out to us at #help on https://reviewpad.com/discord for further assistance.")
+			default:
+				return nil, "", err
+			}
+		}
```
ferreiratiago, line 365: I would suggest setting the `Not found` option as the default one. This is to make sure we cover all cases regardless.

Also, create a common `contactUs` and make the message clear. Here's my suggestion:

```suggestion
		if responseErr, ok := err.(*github.ErrorResponse); ok && responseErr.Response.StatusCode == 404 {
			contactUs := "If you still encounter this error, please reach out to us at #help on https://reviewpad.com/discord for additional support."
			switch responseErr.Message {
			case "Repository access blocked":
				return nil, "", fmt.Errorf("we encountered an authorization error while processing the 'extends' property in your Reviewpad configuration. Please ensure that you have the Reviewpad GitHub App installed in all repositories you are trying to extend from. %s", contactUs)
			default:
				return nil, "", fmt.Errorf("we encountered an error while processing the 'extends' property in your Reviewpad configuration. This problem may be due to incorrect URL or unauthenticated access. Please ensure that you have Reviewpad GitHub App installed in all repositories you are trying to extend from and that the file name and path are correct. %s", contactUs)
			}
		}
```

ferreiratiago (changes requested):

shay2025:
@ferreiratiago I decided to remove the `Repository access blocked`.
If a GitHub App tries to access a repository that it does not have access to, it will receive a "404 Not Found" error instead of the `Repository access blocked` error. This is because the "404 Not Found" error indicates that the requested resource could not be found on the server, whereas the "Repository access blocked" error is a more specific error message for when a user is not allowed to access a repository.

When a GitHub App receives a "404 Not Found" error, it means that the repository does not exist or that the App does not have sufficient permissions to access it.

GitHub Apps do not return the "Repository access block" error because the error message is intended for human users. When a user tries to access a repository that they do not have permission to, they see the "Repository access block" error message in their web browser or in the GitHub UI.

GitHub Apps, on the other hand, communicate with the GitHub API and receive responses in a machine-readable format. The "404 Not Found" error, which indicates that the requested resource could not be found, is a more appropriate error message for GitHub Apps to receive, as it can be easily processed and handled programmatically.

The "Repository access block" error message is intended to provide clear and concise information to human users about why they cannot access a repository, whereas the "404 Not Found" error provides the necessary information for GitHub Apps to determine that the repository is unavailable and to handle the error in an appropriate way.

ferreiratiago (approved):

PR #662, opened 2023-02-01
feat: add integration tests to build

ferreiratiago:
## Description

This pull request introduces the integrations test on both main and pr builds.->

## Type of change

## How was this tested?

Manual tests

## Checklist


integration_test/integration_test.go
```diff
@@ -98,7 +98,7 @@ func TestIntegration(t *testing.T) {
 	subFile, err := utils.ReadFile("./assets/utils/sub.go")
 	require.Nil(err)
 
-	readmeFile, err := utils.ReadFile("./assets/README.md")
+	readmeFile, err := utils.ReadFile("./assets/readme.md")
```
zolamk, line 101: I think we need to rename here https://github.com/reviewpad/reviewpad/blob/e90ed38a49bf65396352ed00115bcb2861605729/integration_test/assets/builtin-others.yml#L85 as well, since we are using it to check `$hasFileName` built-in

PR #666, opened 2023-02-03
feat: separate check for fail built in

### LINKED ISSUE #652: $fail built-in should create a GitHub check
## Context
Currently, the use of the built-in`fail` triggers the execution of Reviewpad to fail. This is confusing because Reviewpad can fail due to an execution error (i.e. GitHub API error) or due to an intentional `fail` defined by the user.

Therefore, when Reviewpad fails is not clear if it failed because there was an internal error or because the user deliberately decided to `fail`.

In addition, if we desire [block the GitHub's merge button](https://docs.reviewpad.com/use-cases/blocking-merge/) using Reviewpad ability to `fail` we disable Reviewpad's ability to auto-merge. This is because we need to make Reviewpad check a mandatory check to enable the merge button (in order to disable the merge button when Reviewpad `fails`.) By making Reviewpad a mandatory check then Reviewpad will never be able to auto-merge because its own check (that is mandatory) is still running by the time it tries to merge.

## Proposed Solution
In order to (1) enable Reviewpad to both block the merge button and be able to do auto-merge and (2) make it clear to the user that a failure happened because the user intentionally failed it we need to update the way the `fail` built-in works.

The proposed solution is the following:
- [ ] The `fail` built-in receives 2 parameters:
   - **Check name**: The name of the check used to identify the check. This is the name that the user will look for when selecting the checks as a [branch protection rule on GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule) to block the merge button.
   - **Reason**: The reason why the failed checked. These are the details that will appear on the check explaining why the check has failed. 
- [ ] The `fail` built-in creates a commit status on the pull request. **This means that each time the `fail` built-in is used a new commit status is created**. For instance, if a Reviewpad configuration has 5 fails and 3 have been triggered, then the pull request will have 3 different commit statuses.
- [ ] The `fail` built-in should **NOT** fail the Reviewpad check.

## Expected Behaviour
After this issue has been solved we should be able to:
- [ ] Use the `fail` built-in to create a commit status with a specific name and description.
- [ ] Use several `fail` built-ins to create several commits status.
- [ ] Use the commit status created by the `fail` built-in as a branch protection rule to block a merge button.
- [ ] Do not affect the Reviewpad check when the `fail` built-in is triggered.

zolamk:
## Description
Currently, the `$fail` built-in fails reviewpad, this PR modifies reviewpad so that `$fail` built-in instead of failing reviewpad creates a new commit status with a `failure` state.

## Related issue

Closes #652 

## Type of change

## How was this tested?
Manual tests performed on the action and on the github app.

## Checklist

### DIFF
--- a/lang/aladino/interpreter.go
+++ b/lang/aladino/interpreter.go
@@ -7,7 +7,6 @@ package aladino
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 
 	gh "github.com/reviewpad/reviewpad/v3/codehost/github"
@@ -17,6 +16,10 @@ import (
 	"github.com/reviewpad/reviewpad/v3/handler"
 	"github.com/reviewpad/reviewpad/v3/utils"
 	"github.com/sirupsen/logrus"
+)
+
+const (
+	commitStatusDescriptionMaxLength = 140
 )
 
 type Interpreter struct {
@@ -112,6 +115,7 @@ func (i *Interpreter) ExecProgram(program *engine.Program) (engine.ExitStatus, e
 	i.Env.GetLogger().Info("executing program")
 
 	failBuiltinStatus := "success"
+	failureReason := "Reviewpad commit status check succeeded."
 
 	for _, statement := range program.GetProgramStatements() {
 		err := i.ExecStatement(statement)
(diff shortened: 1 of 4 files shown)


ferreiratiago:
Closing in favor of executing the proposed solution listed on https://github.com/reviewpad/reviewpad/issues/652

lang/aladino/interpreter.go
```diff
@@ -242,6 +248,27 @@ func (i *Interpreter) ReportMetrics() error {
...
+func (i *Interpreter) createReviewpadFailCommitStatus(state string) error {
+	targetEntity := i.Env.GetTarget().GetTargetEntity()
+
+	if targetEntity.Kind == handler.PullRequest {
+		ctx := i.Env.GetCtx()
+		pr := i.Env.GetTarget().(*target.PullRequestTarget).PullRequest
+		log.Println(pr.GetHead().GetSHA())
+
+		_, err := i.Env.GetGithubClient().CreateCommitStatus(ctx, targetEntity.Owner, targetEntity.Repo, pr.GetHead().GetSHA(), &gh.CreateCommitStatusOptions{
+			Context:     "Reviewpad Fail",
```
ferreiratiago, line 260: Let's rename this to `Reviewpad Commit Status Check`

lang/aladino/interpreter.go
```diff
@@ -242,6 +248,27 @@ func (i *Interpreter) ReportMetrics() error {
...
+			State:       state,
+			Description: "Reviewpad $fail built-in status.",
```
ferreiratiago, line 262: Let's add all fail descriptions separated by commas.

ferreiratiago (approved):

PR #667, opened 2023-02-03
fix: empty reviewpad report

### LINKED ISSUE #654: Reviewpad generates empty report
**Describe the bug**
When Reviewpad is configured in `mode: verbose` it generates a report after each run.

However, if the execution has no `Executed Actions` Reviewpad generates an empty report.

**To Reproduce**
1. Configure Reviewpad with `mode: verbose` and in a way that it is executed by no actions are triggered
2. Open a pull request to trigger Reviewpad to run
3. Reviewpad runs and an empty report is put in place.

Screen Shot 2023-01-31 at 10 15 54 AM

**Expected behavior**
We should make it clear that no actions were executed using the text `no actions executed`.

Screen Shot 2023-01-31 at 10 14 42 AM

shay2025:
## Description

This pull request resolves the issue where the Reviewpad generates an empty report when the `mode` property is set to `verbose`. The solution handles the scenario where there are no actions executed by generating a report indicating this, otherwise, it presents a list of the actions executed.

## Related issue

Closes #654 

## Type of change

## How was this tested?

Manual tests (described here https://www.notion.so/reviewpad/Reviewpad-generates-an-empty-report-e5eb16579ae84e2ab10f4464241a2dc9) and ran the `task check -f` cmd.

## Checklist

### DIFF
--- a/lang/aladino/report.go
+++ b/lang/aladino/report.go
@@ -90,8 +90,12 @@ func BuildVerboseReport(report *Report) string {
 	sb.WriteString("```yaml\n")
 
 	// Report
-	for _, action := range report.Actions {
-		sb.WriteString(fmt.Sprintf("%v\n", action))
+	if len(report.Actions) == 0 {
+		sb.WriteString("No actions executed\n")
+	} else {
+		for _, action := range report.Actions {
+			sb.WriteString(fmt.Sprintf("%v\n", action))
+		}
 	}
 
 	sb.WriteString("```\n")
--- a/lang/aladino/interpreter_internal_test.go
+++ b/lang/aladino/interpreter_internal_test.go
@@ -620,7 +620,7 @@ func TestReport_OnSilentMode_WhenNoReviewpadCommentIsFound(t *testing.T) {
 
 func TestReport_OnVerboseMode_WhenNoReviewpadCommentIsFound(t *testing.T) {
 	var addedComment string
-	commentToBeAdded := fmt.Sprintf("%s%s\n**Reviewpad Report**\n\n:scroll: **Executed actions**\n```yaml\n```\n", ReviewpadReportCommentAnnotation, ReviewpadIgnoreCommentAnnotation)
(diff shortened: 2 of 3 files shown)


zolamk (approved):
LGTM

lang/aladino/report.go
```diff
@@ -90,8 +90,12 @@ func BuildVerboseReport(report *Report) string {
 	sb.WriteString("```yaml\n")
 
 	// Report
-	for _, action := range report.Actions {
-		sb.WriteString(fmt.Sprintf("%v\n", action))
+	if len(report.Actions) == 0 {
+		sb.WriteString("No actions executed\n")
+	} else {
```
ferreiratiago, line 95: We can return immediately here and allow for better reading.

```suggestion
	if len(program.GetProgramStatements()) == 0 {
		sb.WriteString("No actions triggered\n```\n")
		return sb.String()
	}
```

ferreiratiago (changes requested):
Minor improvement.

lang/aladino/report.go
shay2025: Hmm, how will this allow for better reading, @ferreiratiago?
We are only handling the case of whether we have actions executed or not, after that the treatment is exactly the same.
I don't see the need to repeat the code unnecessarily 🤔

ferreiratiago: It's the idea of [line of sight](https://speakerdeck.com/fedepaol/reducing-cognitive-load-yet-another-idiomatic-go-talk?slide=20) in order to reduce the cognitive load. In this case in particular is not a big deal but I think we should take this idea into account when making changes. You can check Federico [talk](https://fosdem.org/2023/schedule/event/goreducecognitive/) here.

shay2025: @ferreiratiago Yes, I am aware of the concept precept. However, being this such a simple case, adding more code may do the exact opposite and cause noise.

ferreiratiago (approved):

PR #685, opened 2023-02-10
feat: approve built-in

### LINKED ISSUE #677: Define built-in approve
## Context
We should be able to approve a pull request.

## Proposed Solution
Create a new built-in called `approve`.

This built-in should be able to approve a pull request.

By default, this action is only enabled for open pull requests that are not draft.

This action is a shortcut of `$review("APPROVE", "")`.

### Parameters

| --- | --- | --- |
| `comment` (optional) | string | The body of the comment |

### Example
`$approve("This PR contains only trivial changes")`

shay2025:
## Description

This pull request introduces the `$approve` built-in with the purpose of performing an approval to the pull request with or without a comment.

## Related issue

Closes #677 

## Type of change

## How was this tested?

Manual tests described in https://www.notion.so/reviewpad/approve-built-in-47c4537d625e4f1db50726766326e3e5 and ran the `task check -f` command. 

## Checklist


plugins/aladino/actions/approve.go
```diff
@@ -0,0 +1,69 @@
...
+
+func Approve() *aladino.BuiltInAction {
+	return &aladino.BuiltInAction{
+		Type:           aladino.BuildFunctionType([]aladino.Type{aladino.BuildStringType()}, nil),
+		Code:           approveCode,
+		SupportedKinds: []handler.TargetEntityKind{handler.PullRequest},
+	}
+}
+
+func approveCode(e aladino.Env, args []aladino.Value) error {
```
ferreiratiago, line 21: Does it make sense to make this `approve` function a shortcut to `review`, i.e.

```go
func approveCode(e aladino.Env, args []aladino.Value) error {
	return reviewCode(e, []aladino.Value{
		&aladino.StringValue{Val: "APPROVE"},
		args[0],
	})
}
````

This way we avoid repeating all logic for such as:
- when pr is closed
- no changes
- ...

Logic, such as when draft we could move it into the `review` builtin. We can ask Freddy to check if that's ok.

WDYT?

File: plugins/aladino/actions/approve.go [hunk 2]
```diff
@@ -0,0 +1,69 @@
...
+			log.Infof("skipping approve because there were no updates since the last review")
+			return nil
+		}
+
+		log.Infof("latest review from %v is %v with body '%v'", authenticatedUserLogin, latestReview.State, latestReview.Body)
+
+		if latestReview.State == "APPROVED" && latestReview.Body == reviewBody {
+			log.Infof("skipping approve since it's the same as the latest review")
+			return nil
+		}
```
ferreiratiago, line 64: After discussing with @fmallet we believe that this condition should be removed. If there are any changes, Reviewpad should approve again the pull request.

ferreiratiago: We can add this suggestion to the `review` built-in.

shay2025: @ferreiratiago In that case, how do you think we should proceed?
1. should we remove that condition from the `$approve` built-in and accept it
2. should we still apply the suggestion in https://github.com/reviewpad/reviewpad/pull/685#discussion_r1104621600 and afterward handle the `review` built-in
3. should we change the `review` built-in with the removal of that condition and apply the suggestion in https://github.com/reviewpad/reviewpad/pull/685#discussion_r1104621600

ferreiratiago: I believe option 3 is the best :)

plugins/aladino/actions/approve_test.go
```diff
@@ -0,0 +1,383 @@
...
+	host "github.com/reviewpad/reviewpad/v3/codehost/github"
+	"github.com/reviewpad/reviewpad/v3/lang/aladino"
+	plugins_aladino "github.com/reviewpad/reviewpad/v3/plugins/aladino"
+	"github.com/reviewpad/reviewpad/v3/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+var approve = plugins_aladino.PluginBuiltIns().Actions["approve"].Code
+
+func TestApprove(t *testing.T) {
```
ferreiratiago, line 25: Shouldn't these tests be under `review` builtin?

plugins/aladino/actions/review.go
```diff
@@ -26,11 +22,15 @@ func Review() *aladino.BuiltInAction {
...
 func reviewCode(e aladino.Env, args []aladino.Value) error {
 	t := e.GetTarget().(*target.PullRequestTarget)
-	entity := e.GetTarget().GetTargetEntity()
-	clientGraphQL := e.GetGithubClient().GetClientGraphQL()
+
 	log := e.GetLogger().WithField("builtin", "review")
 
-	if *t.PullRequest.State == "closed" {
+	if t.PullRequest.GetDraft() {
+		log.Infof("skipping approve because the pull request is in draft")
```
ferreiratiago, line 29:
```suggestion
		log.Infof("skipping review because the pull request is in draft")
```

plugins/aladino/actions/approve_test.go
ferreiratiago: Since this built-in is a shortcut for `review` all required tests should be in the `review` built-in.

shay2025: Then, I will remove these tests.

ferreiratiago (approved):

PR #689, opened 2023-02-11
feat: compute git blame rank

### LINKED ISSUE #681: Define ComputeGitBlameRank
## Context
Assuming that we are able to run [git blame](https://github.com/reviewpad/reviewpad/issues/679) on a set of files for a specific commit SHA, able to [slice the git blame](https://github.com/reviewpad/reviewpad/issues/680), we now need to able to compute a git blame rank in order to have the list of authors ordered by rank.

## Dependencies
- https://github.com/reviewpad/reviewpad/issues/679
- https://github.com/reviewpad/reviewpad/issues/680

## Proposed Solution
Define a function following the syntax:
`ComputeGitBlameRank(gitBlame: GitBlame) []GitBlameAuthorRank`

### Parameters
- `gitBlame` is a calculated GitBlame

### Return value
A new list of `GitBlameAuthorRank` order by descending order based on the total number of lines that a user is an author of.

```go
type GitBlameAuthorRank struct {
	Username   string
	TotalLines int
}
```

### Example
```go
sample := GitBlame{
	CommitSHA: "1234567890",
	Files: map[string]GitBlameFile{
		"bar.go": {
			FilePath:  "bar.go",
			LineCount: 10,
			Lines: []GitBlameFileLines{
				{
					FromLine: 1,
					ToLine:   10,
					Author:  "peter",
				},
				{
					FromLine: 11,
					ToLine:   50,
					Author: "maria"
				},
				{
					FromLine: 51,
					ToLine:   52,
					Author: "john",
				}
			},
		},
		"foo.go": {
			FilePath:  "foo.go",
			LineCount: 10,
			Lines: []GitBlameFileLines{
				{
					FromLine: 1,
					ToLine:   15,
					Author: "john",
				}
			}
		},
	},
}

ComputeGitBlameRank(sample)

// OUTPUT
// []{
// 	{
// 		Username: "maria",
// 		TotalLines: 40
// 	},
// 	{
// 		Username: "john",
// 		TotalLines: 17
// 	},
// 	{
// 		Username: "peter",
// 		TotalLines: 10
// 	}
// }
```

zolamk:
## Description
Adds a `ComputeGitBlameRank` function to github client.

## Related issue

Closes #681 

## Type of change

## How was this tested?
Unit Tests

## Checklist


codehost/github/blame.go
```diff
@@ -201,3 +207,34 @@ func (c *GithubClient) SliceGitBlame(blame *GitBlame, filePath string, slices []
...
+
+			authorIndex, ok := authorIndices[author]
+			if !ok {
+				authorIndex = len(rankedAuthors)
+				authorIndices[author] = authorIndex
+				rankedAuthors = append(rankedAuthors, GitBlameAuthorRank{
+					Username:   author,
+					TotalLines: lineCount,
+				})
+				continue
```
ferreiratiago, line 228:
```suggestion
				rankedAuthors = append(rankedAuthors, GitBlameAuthorRank{
					Username:   author,
					TotalLines: 0,
				})
```

codehost/github/blame.go
```diff
@@ -201,3 +207,34 @@ func (c *GithubClient) SliceGitBlame(blame *GitBlame, filePath string, slices []
 
 	return slicedBlame, nil
 }
+
+func (c *GithubClient) ComputeGitBlameRank(gitBlame *GitBlame) []GitBlameAuthorRank {
+	var rankedAuthors []GitBlameAuthorRank
+	authorIndices := make(map[string]int)
```
ferreiratiago, line 213: WDYT about using a map to gather authors' total lines and then convert them into an array and order it?

```suggestion
	totalLinesByAuthor := make(map[string]int)
```

zolamk: The thing is to do that we would need to have two loops first one to gather them and then another loop to convert them into an array, what do you think? should we separate the steps for more clarity

codehost/github/blame.go
zolamk: when this is the first line range for the author we add the author to the rank list with the number of lines owned and then continue with the rest, is my logic wrong, what do you think?

ferreiratiago: You're logic is fine. If we add the `TotalLines: 0` we can then remove the `continue` and the operation of adding `lineCount` will be done in line [155](https://github.com/reviewpad/reviewpad/pull/689/files#diff-60f7a7344d613b497c5c059d6c187bd1fcf2af62261877d4d61dc7d7049f64feR155). Does this makes sense?

PR #713, opened 2023-02-17
feat: totalCodeReviewsInOrganization built-in

### LINKED ISSUE #701: Define built-in totalCodeReviewsInOrganization
## Context
We should be able to know the total number of code reviews a user has within the organization. Currently, the built-in `totalCodeReviews` only provides the total amount of code reviews within a repository.

## Proposed Solution
Define the built-in `totalCodeReviewsInOrganization` that gets the total number of open code reviews a user has created within an organization.

The organization is the one where the built-in is running.

The total number of code reviews should only count those that are open.

### Parameters

| --- | --- | --- |
| `user_login` | string | The GitHub user login. |

### Example
`$totalCodeReviewsInOrganization("peter")`

shay2025:
## Description

This pull request introduces the `$totalCodeReviewsInOrganization` built-in.

The built-in will only count the code reviews from open pull requests that live in repositories where the Reviewpad GitHub App is installed.

## Related issue

Closes #701 

## Type of change

## How was this tested?

The manual tests are described in https://www.notion.so/reviewpad/totalCodeReviewsInOrganization-built-in-6fcdd71f813147d2a5c1f0e45e1944c5

## Checklist

### DIFF
--- a/codehost/github/reviews_test.go
+++ b/codehost/github/reviews_test.go
@@ -42,7 +42,7 @@ func TestGetReviewsCountByUserFromOpenPullRequests(t *testing.T) {
 	tests := map[string]struct {
 		ghGraphQLHandler func(http.ResponseWriter, *http.Request)
 		wantCount        int
-		wantErr          string
+		wantErr          error
 	}{
 		"when the request for the reviews count fails": {
 			ghGraphQLHandler: func(w http.ResponseWriter, req *http.Request) {
@@ -51,7 +51,7 @@ func TestGetReviewsCountByUserFromOpenPullRequests(t *testing.T) {
 					http.Error(w, "GetReviewsCountFail", http.StatusNotFound)
 				}
 			},
-			wantErr: "non-200 OK status code: 404 Not Found body: \"GetReviewsCountFail\\n\"",
+			wantErr: fmt.Errorf("non-200 OK status code: 404 Not Found body: \"GetReviewsCountFail\\n\""),
 		},
 		"when the user has created reviews in open pull requests": {
 			ghGraphQLHandler: func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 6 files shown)


codehost/github/reviews_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			gotCount, gotErr := mockedEnv.GetGithubClient().GetReviewsCountByUserFromOpenPullRequests(mockedEnv.GetCtx(), aladino.DefaultMockPrOwner, "test")
+
+			if test.wantErr != "" {
+				assert.EqualError(t, gotErr, test.wantErr)
+				assert.Equal(t, 0, gotCount)
+			} else {
+				assert.Equal(t, test.wantCount, gotCount)
+				assert.Nil(t, gotErr)
+			}
```
zolamk, line 102: Do you think it might be better to make `wantErr` an error and assert equal
```suggestion
			assert.Equal(t, test.wantErr, gotErr)
			assert.Equal(t, test.wantCount, gotCount)
```

plugins/aladino/functions/totalCodeReviewsInOrganization.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func totalCodeReviewsInOrganizationCode(e aladino.Env, args []aladino.Value) (aladino.Value, error) {
+	username := args[0].(*aladino.StringValue).Val
+	entity := e.GetTarget().GetTargetEntity()
+
+	ghPullRequest, _, err := e.GetGithubClient().GetPullRequest(e.GetCtx(), entity.Owner, entity.Repo, entity.Number)
+	if err != nil {
+		return nil, err
+	}
```
zolamk, line 30: why do we need to get the PR from the API, won't the PR from the env work, what do you think?

shay2025: I don't believe the PR in the env work, as it is right now, is enough. We need to get the owner of the repo where the pull request is. This info is not available in the PR of the env.

zolamk: hmm, a little confusing it wouldn't have that info because the PR in the env also calls the `GetPullRequest` method on the github client right https://github.com/reviewpad/reviewpad/blob/081ab53bac20bbc500fc289fd48b8d4ff4a18b0c/lang/aladino/env.go#L151 or am I missing something

shay2025: You are right 👌

ferreiratiago:
Closing due to bad specification.

PR #714, opened 2023-02-19
feat: improve random reviewer selection

### LINKED ISSUE #709: Tune the random selection step of $assignCodeAuthorReviewers() when no author available
The selection algorithm implemented in `$assignCodeAuthorReviewers()` is the following:
1. Take the list of active collaborators on this GitHub repository
2. In this list, filter out the ones whose usernames contain [bot]
3. In this list, extract the ones who wrote some lines of code in the latest version of the impacted source files
4. Sort those authors according to the number of written lines of code (on the whole files)
5. Starting from the top of this author list, take the first one having less than three open PRs (configurable) in review
6. If we end up with no author, randomly pick a reviewer in the list of active collaborators
7. Assign this reviewer to the PR

If we reach step 6, we can select an author part of step 5, so a reviewer with more open PRs than the expected threshold. We are missing the following intermediate step between 5 and 6 to improve the behavior slightly:

5.5. If we end up with no author, take the full list of active collaborators on this project, exclude from this list the authors used in step 5, and randomly pick the requested number of reviewers having less than the threshold number of open PRs in review.

zolamk:
## Description
Improves the random reviewer selection for the `$assignCodeAuthorReviewers` built-in function.

## Related issue

Closes #709 

## Type of change

## How was this tested?
Unit and manual tests

## Checklist

### DIFF
--- a/plugins/aladino/actions/assignCodeAuthorReviewers.go
+++ b/plugins/aladino/actions/assignCodeAuthorReviewers.go
@@ -7,6 +7,7 @@ package plugins_aladino_actions
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"strings"
 
 	gh "github.com/google/go-github/v49/github"
@@ -15,7 +16,6 @@ import (
 	"github.com/reviewpad/reviewpad/v3/codehost/github/target"
 	"github.com/reviewpad/reviewpad/v3/handler"
 	"github.com/reviewpad/reviewpad/v3/lang/aladino"
-	"github.com/reviewpad/reviewpad/v3/utils"
 	"golang.org/x/exp/slices"
 )
 
@@ -64,7 +64,7 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error 
 	}
 
 	// Get all available assignees that are not authors
-	nonAuthorAvailableAssignees := getNonAuthorAvailableAssignees(availableAssignees, authors)
+	nonAuthorAvailableAssignees := filterAuthorsFromAssignees(availableAssignees, authors)
 
 	// Fetch the total number of open pull requests that each author has already assigned to them as a reviewer.
 	totalOpenPRsAsReviewerByUser, err := gitHubClient.GetOpenPullRequestsAsReviewer(env.GetCtx(), pr.GetTargetEntity().Owner, pr.GetTargetEntity().Repo, append(authors, nonAuthorAvailableAssignees...))
@@ -83,7 +83,7 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error 
 	// Find eligible reviewers from the available assignees and pick one at random
 	if len(selectedReviewers) == 0 {
 		for _, assignee := range nonAuthorAvailableAssignees {
-			if isAssigneeEligibleToReview(assignee, pr.PullRequest, reviewersToExclude, totalOpenPRsAsReviewerByUser[assignee], maxAllowedAssignedReviews) {
+			if isUserEligibleToReview(assignee, pr.PullRequest, reviewersToExclude, availableAssignees, totalOpenPRsAsReviewerByUser[assignee], maxAllowedAssignedReviews) {
 				selectedReviewers = append(selectedReviewers, assignee)
 			}
 		}
@@ -94,9 +94,7 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error 
 			return assignRandomReviewerCode(env, nil)
 		}
 
-		randomReviewer := selectedReviewers[utils.GenerateRandom(len(selectedReviewers))]
-
-		return pr.RequestReviewers([]string{randomReviewer})
+		selectedReviewers = []string{selectedReviewers[rand.Intn(len(selectedReviewers))]}
 	}
 
 	if totalRequiredReviewers > len(selectedReviewers) {
@@ -110,7 +108,7 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error 
 	return pr.RequestReviewers(selectedReviewers)
 }
 
-func getNonAuthorAvailableAssignees(availableAssignees []*codehost.User, authors []string) []string {
+func filterAuthorsFromAssignees(availableAssignees []*codehost.User, authors []string) []string {
 	assignees := []string{}
 	for _, assignee := range availableAssignees {
 		if !slices.Contains(authors, assignee.Login) {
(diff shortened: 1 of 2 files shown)


plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -89,6 +110,17 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error
...
+func getNonAuthorAvailableAssignees(availableAssignees []*codehost.User, authors []string) []string {
+	assignees := []string{}
+	for _, assignee := range availableAssignees {
+		if !slices.Contains(authors, assignee.Login) {
+			assignees = append(assignees, assignee.Login)
+		}
+	}
+
+	return assignees
+}
```
ferreiratiago, line 122: I would suggest to rename this function to `func filterAuthorsFromAssignees(assignees []*codehost.User, authors []string) []string {`. WDTY?

plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -111,11 +143,10 @@ func getAuthorsFromGitBlame(ctx context.Context, gitHubClient *github.GithubClie
 	return authors, nil
 }
 
-func isUserEligibleToReview(
+func isAssigneeEligibleToReview(
```
ferreiratiago, line 146: I understand the reason for this function. I would suggest keeping the function `isUserEligibleToReview` as it was and calling it also filtering the eligible users when assignee.

plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -74,8 +79,24 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error
...
+		// If we couldn't find any eligible reviewers from the available assignees that are not authors
+		// Assign a random reviewer
+		if len(selectedReviewers) == 0 {
+			return assignRandomReviewerCode(env, nil)
+		}
+
+		randomReviewer := selectedReviewers[utils.GenerateRandom(len(selectedReviewers))]
+
+		return pr.RequestReviewers([]string{randomReviewer})
 	}
```
ferreiratiago, line 100: I think we can improve this code for better understanding.
1. If there's no `selectedReviewers` then try to get then from `nonAuthorAvailableAssignees`.

```go
	if len(selectedReviewers) == 0 {
		for _, author := range nonAuthorAvailableAssignees {
			if isUserEligibleToReview(author, pr.PullRequest, reviewersToExclude, availableAssignees, totalOpenPRsAsReviewerByUser[author], maxAllowedAssignedReviews) {
				selectedReviewers = append(selectedReviewers, author)
			}
		}

		selectedReviewers = []string{selectedReviewers[rand.Intn(len(selectedReviewers))]}
	}
```

2. If there's no `selectedReviewers` then assign random reviewer.

```go
	if len(selectedReviewers) == 0 {
		return assignRandomReviewerCode(env, nil)
	}
```

WDYT?

plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -74,8 +79,24 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error
...
+			}
+		}
+
+		// If we couldn't find any eligible reviewers from the available assignees that are not authors
+		// Assign a random reviewer
+		if len(selectedReviewers) == 0 {
+			return assignRandomReviewerCode(env, nil)
+		}
+
+		randomReviewer := selectedReviewers[utils.GenerateRandom(len(selectedReviewers))]
```
ferreiratiago, line 97: We should be using `rand.Intn` instead of `utils.GenerateRandom`.

ferreiratiago (changes requested):

plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -159,3 +189,18 @@ func isUserValidAssignee(assignees []*codehost.User, username string) bool {
...
+
+	randomReviewers := make([]string, requiredNumberOfReviewers)
+	for i := 0; i < requiredNumberOfReviewers; i++ {
+		randIndex := rand.Intn(len(reviewers))
+		randomReviewers[i] = reviewers[randIndex]
+		reviewers = append(reviewers[:randIndex], reviewers[randIndex+1:]...)
+	}
+
+	return randomReviewers
+}
```
ferreiratiago, line 206: Suggestion for better reading.

```suggestion
func getRandomReviewers(reviewers []string, total int) []string {
	if total >= len(reviewers) {
		return reviewers
	}

	selectedReviewers := make([]string, total)

	for i := 0; i < total; i++ {
		randIndex := rand.Intn(len(reviewers))
		selectedReviewers[i] = reviewers[randIndex]

		reviewers = append(reviewers[:randIndex], reviewers[randIndex+1:]...)
	}

	return selectedReviewers
}
```

plugins/aladino/actions/assignCodeAuthorReviewers.go
```diff
@@ -74,12 +79,26 @@ func assignCodeAuthorReviewersCode(env aladino.Env, args []aladino.Value) error
...
+			if isUserEligibleToReview(assignee, pr.PullRequest, reviewersToExclude, availableAssignees, totalOpenPRsAsReviewerByUser[assignee], maxAllowedAssignedReviews) {
+				selectedReviewers = append(selectedReviewers, assignee)
+			}
+		}
+
+		// If we couldn't find any eligible reviewers from the available assignees that are not authors
+		// Assign a random reviewer
+		if len(selectedReviewers) == 0 {
+			return assignRandomReviewerCode(env, nil)
+		}
```
ferreiratiago, line 95: My understanding may be wrong but if we are not able to assign random reviewers from the available assignees why would `assignRandomReviewerCode` work since it does the same? 😄 I think we can remove this piece of code.

ferreiratiago (changes requested):

plugins/aladino/actions/assignCodeAuthorReviewers.go
zolamk: The thing is if none of the code authors are eligible then we try to find eligible reviewers from the available assignees if that still doesn't give us any reviewers then we pick a random reviewer regardless of their eligibility right, I think it's step 6 of the specification, what do you think?

zolamk: the `$assignRandomReviewer` built-in doesn't have any eligibility check so it will just pick a random user right

ferreiratiago: Indeed @zolamk That is correct :)

ferreiratiago (approved):

PR #724, opened 2023-02-23
feat: prepare release

ferreiratiago:
## Description

This pull request introduces a series of fixes for the release.

## Related issue

_none_

## Type of change

## Checklist

### DIFF
--- a/plugins/aladino/actions/review.go
+++ b/plugins/aladino/actions/review.go
@@ -61,19 +61,21 @@ func reviewCode(e aladino.Env, args []aladino.Value) error {
 	}
 
 	if latestReview != nil {
-		// The last push was made before the last review so a new review is not needed
-		if lastPushDate.Before(*latestReview.SubmittedAt) {
-			log.Infof("skipping review because there were no updates since the last review")
-			return nil
-		}
-
 		latestReviewEvent, err := mapReviewStateToEvent(latestReview.State)
 		if err != nil {
 			return err
 		}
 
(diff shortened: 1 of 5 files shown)


plugins/aladino/actions/review.go
```diff
@@ -61,19 +61,21 @@ func reviewCode(e aladino.Env, args []aladino.Value) error {
...
-		}
-
 		latestReviewEvent, err := mapReviewStateToEvent(latestReview.State)
 		if err != nil {
 			return err
 		}
 
+		// If the latest review is the same as the one we want to create, and the last push date is before the latest review
+		// then we skip the review creation.
+		if latestReview.State == reviewEvent && latestReview.Body == reviewBody && lastPushDate.Before(*latestReview.SubmittedAt) {
```
zolamk, line 71: This condition will never be true because we are not mapping the state to the event right
```suggestion
		if latestReviewEvent == reviewEvent && latestReview.Body == reviewBody && lastPushDate.Before(*latestReview.SubmittedAt) {
```

ferreiratiago: I'm using `latestReviewEvent` which is a result from `mapReviewStateToEvent(latestReview.State)`. Isn't this what you're saying?

zolamk: I already pushed a fix for this review :sweat_smile: , just wanted to leave the reason for the change here, it should be like that right

PR #729, opened 2023-02-27
chore: load reviewpad file from gh method

shay2025:
## Description

This pull request transfers the methods that load the reviewpad file contents from GitHub and check if the pull request changes the reviewpad file which lived in the GitHub Action.

## Type of change

**Improvements** (non-breaking change without functionality)

## How was this tested?

Manual tests in the dev app where 2 pull requests were opened to test when:
- the changes contain the reviewpad file;
- the changes do not contain the reviewpad file.

## Checklist

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -52,6 +52,17 @@ func Load(ctx context.Context, log *logrus.Entry, githubClient *gh.GithubClient,
 	return file, nil
 }
 
+func LoadReviewpadFileFromGitHub(ctx context.Context, logger *logrus.Entry, githubClient *gh.GithubClient, filePath string, branch *github.PullRequestBranch) (*engine.ReviewpadFile, error) {
+	reviewpadFileContent, err := githubClient.DownloadContents(ctx, filePath, branch)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


runner.go
```diff
@@ -52,6 +52,17 @@ func Load(ctx context.Context, log *logrus.Entry, githubClient *gh.GithubClient,
 	return file, nil
 }
 
+func LoadReviewpadFileFromGitHub(ctx context.Context, logger *logrus.Entry, githubClient *gh.GithubClient, filePath string, branch *github.PullRequestBranch) (*engine.ReviewpadFile, error) {
```
ferreiratiago, line 55: I would suggest moving this into `utils/file.go`

shay2025: Hmm, it might not be the best place to go because that would cause an import cycle since I would need to import the `reviewpad` package into the `utils` package because of the [Load](https://github.com/reviewpad/reviewpad/pull/729/files#diff-0ccec778a288e27b5a6ada45110d2a1855585629224a3dd4a8b5ff6f9319e283R63) function

ferreiratiago (approved):

PR #737, opened 2023-02-28
feat: select amount of desired assignees in assignAssignees

### LINKED ISSUE #733: Update built-in assignAssignees
### WHY

When the input parameter of the `$assignAssignees(...)` built-in is a group, there is no way to assign the Pull Request to a limited number of persons. We need to be able to assign a limited number of assignees from the provided list.

### Motivation

This change will allow us to select a set of assignees from a group.

```yml
groups:
  - name: owners
    spec: '["maria", "john"]'

workflows:
  - name: assign
    always-run: true
    if:
      - rule: '!$isDraft()'
        extra-actions:
          - $assignAssignees($group("owners"), 1)
```

In this case, we are able to assign the pull request to only 1 user from the group `owners`.

## Proposed solution

Add a new parameter called `total` with the total number of users to assign the pull request to.

### Update parameters

| --- | --- | --- |
| assignees | []string | The list of GitHub logins to assign the pull request / issue to. |
| total (optional) | int | The total number of users to assign to. By default, it assigns to all users in the provided list. | 

### Example
`$assignAssignees($group("owners"), 1)`

shay2025:
## Description

This pull request updates the `$assignAssignees` built-in to include another parameter, which is optional, that defines the total number of users to be assigned.
If the introduced parameter is not set, the default value is the total number of given users (the 1st parameter) if the total does not exceed 10. To achieve this, in the [`transform`](https://github.com/reviewpad/reviewpad/blob/main/engine/transform.go) process we set a `REVIEWPAD_DEFAULT_INT_VALUE` which is `-1` to be the default of the total required users to be assigned and on the execution of the built-in once we detect this tag we know that the total required users to be assigned should be the number of given users that can be assigned.

## Related issue

Closes #733

## Type of change

## How was this tested?

Manual tests were performed in the dev app where the test cases were:
- `$assignAssignees(["dummypad"])` ✅ 
- `$assignAssignees(["dummypad"], 2)` ✅
- `$assignAssignees(["dummypad"], -3)` ❌ `assignAssignees: total required assignees is invalid. please insert a number bigger than 0.`
- `$assignAssignees($group("two-dummies"))` ✅ 
- `$assignAssignees($group("two-dummies"), 1)` ✅ 
- `$assignAssignees($team("engineering"))` ✅ 
- `$assignAssignees($team("engineering"), 2)` ✅ 

## Checklist

### DIFF
--- a/engine/transform.go
+++ b/engine/transform.go
@@ -122,7 +122,7 @@ func addDefaultHasAnyCheckRunCompleted(str string) string {
 
 func addDefaultsToRequestedAssignees(str string) string {
 	trimmedStr := strings.ReplaceAll(str, " ", "")
-	m := regexp.MustCompile(`\$assignAssignees\(((\[.*\])|(\$(team|group)\("[^"]*"\)))(,(-)?(\d+))?\)`)
+	m := regexp.MustCompile(`\$assignAssignees\(((\[.*\])|(\$(team|group)\("[^"]*"\)))(,((-)?(\d+)))?\)`)
 	match := m.FindStringSubmatch(trimmedStr)
 
 	if len(match) == 0 {
--- a/engine/transform.go
+++ b/engine/transform.go
@@ -122,7 +122,7 @@ func addDefaultHasAnyCheckRunCompleted(str string) string {
 
 func addDefaultsToRequestedAssignees(str string) string {
 	trimmedStr := strings.ReplaceAll(str, " ", "")
-	m := regexp.MustCompile(`\$assignAssignees\(((\[.*\])|(\$(team|group)\("[^"]*"\)))(,(\d+))?\)`)
+	m := regexp.MustCompile(`\$assignAssignees\(((\[.*\])|(\$(team|group)\("[^"]*"\)))(,(-)?(\d+))?\)`)
 	match := m.FindStringSubmatch(trimmedStr)
 
 	if len(match) == 0 {
--- a/engine/transform.go
+++ b/engine/transform.go
@@ -141,7 +141,7 @@ func addDefaultsToRequestedAssignees(str string) string {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


plugins/aladino/actions/assignAssignees.go
```diff
@@ -6,34 +6,95 @@ package plugins_aladino_actions
...
+	totalRequiredAssignees := args[1].(*aladino.IntValue).Val
+	log := e.GetLogger().WithField("builtin", "assignAssignees")
+
+	if len(availableAssignees) == 0 {
 		return fmt.Errorf("assignAssignees: list of assignees can't be empty")
 	}
 
-	if len(assignees) > 10 {
+	if totalRequiredAssignees == 0 {
+		return fmt.Errorf("assignAssignees: total required assignees is invalid. please insert a number bigger than 0.")
```
zolamk, line 37: should we use the logger for this

engine/transform.go
```diff
@@ -6,9 +6,14 @@ package engine
 
 import (
 	"regexp"
+	"strconv"
 	"strings"
 )
 
+const (
+	REVIEWPAD_DEFAULT_INT_VALUE string = "-1"
```
zolamk, line 14: could we make this into an `int` instead of converting it, what do you think?

plugins/aladino/actions/assignAssignees.go
```diff
@@ -6,34 +6,95 @@ package plugins_aladino_actions
...
+	}
+
+	reviewpadDefaultIntValue, err := strconv.Atoi(engine.REVIEWPAD_DEFAULT_INT_VALUE)
+	if err != nil {
+		return err
```
zolamk, line 42: we could wrap the errors with a descriptive message, what do you think?

ferreiratiago (approved):

PR #761, opened 2023-03-07
fix: consider last user review in hasRequiredApprovals

### LINKED ISSUE #754: Built-in hasRequiredApprovals ignores latest reviews
## Describe the bug

Currently, the built-in [`hasRequiredApprovals`](https://docs.reviewpad.com/guides/built-ins/#hasrequiredapprovals) returns `true` if one of the required approvals has an approval review regardless of the latest review status.

For instance, if we use the built-in `hasRequiredApprovals(1, ["peter"])` and `peter` does the reviews:
- R1: `Approval`
- R2: `Request changes`

The built-in will return `true` which is incorrect because the latest review from `peter` is `Request changes`.

You can see a live example here https://github.com/reviewpad/schemas/pull/11

## How to reproduce

1. Define a reviewpad.yml with the following configuration (replace `peter` by your GitHub user):

```yml
api-version: reviewpad.com/v3.x

workflows:
  - name: hasRequiredApprovals
    if:
      - $hasRequiredApprovals(1, ["peter"])
    then:
      - $comment("hasRequiredApprovals is true")
```

2. Open a pull request with a user different from the one defined in the configuration
3. Approve the pull request. You will see the comment `hasRequiredApprovals is true`
4. Request changes. You will see the comment `hasRequiredApprovals is true`. This shouldn't be the case.

## Expected behavior

The built-in `hasRequiredApprovals` should look into the latest reviews.

shay2025:
## Description

The `hasRequiredApprovals` built-in was considering if any of the given users had made an approval when it should be checking if the LAST review from any of the given users was an approval.
For this, I inserted a new method `GetApprovedReviewers` which gets the list of reviewers whose last review was approved.
This method first sorts the pull request reviews from latest to oldest and grabs the latest review made by the user and stores it in a map that says if it was approved or not.
After this, we get the list of user logins who approved the review through this map.

## Related issue

Closes #754 

## Type of change

**Bug fix** (non-breaking change which fixes an issue)

## How was this tested?

Manual tests in the dev app where a user made the following reviews (by the order presented):
- review with an approved state;
- review with requested changes.
This test should say the required approvals were not met.

## Checklist


marcelosousa:
/reviewpad robin please summarize these changes. Also, provide a file by file description of the changes and suggest an approval (LGTM / Requires Changes) status.

marcelosousa:
/reviewpad robin please summarize these changes with a summary paragraph and a file by file description of the changes together with an approval status.

marcelosousa:
/reviewpad robin please summarize these changes, provide a file by file description of the changes and suggest an approval status.

ferreiratiago:
/reviewpad robin please summarize these changes, provide a file by file description of the changes and suggest an approval status.

codehost/github/target/pull_request_target.go
```diff
@@ -539,3 +540,33 @@ func (target *PullRequestTarget) GetProjectV2ItemID(projectID string) (string, e
 
 	return target.githubClient.GetPullRequestProjectV2ItemID(ctx, owner, repo, projectID, targetEntity.Number)
 }
+
+func (target *PullRequestTarget) GetApprovedReviewers() ([]string, error) {
```
zolamk, line 544: @shay2025 we can get the latest reviews per reviewer using a query like If It's helpful
```
query { 
  repository(owner: "reviewpad", name: "schemas") {
    pullRequest(number: 11) {
      latestReviews(last: 10) {
        nodes{
          state
          author {
            login
          }
        }
      }
    }
  }
}
```

shay2025: Yes, but first we need to get the reviewers which I use the method `GetReviews` to do so. And to avoid doing more requests I use the reviews as well.

shay2025: Besides that, we already have a method to get the [latest review from a user](https://github.com/reviewpad/reviewpad/blob/main/codehost/github/target/pull_request_target.go#L483). I could use it as well but again I was trying to minimize the number of requests.

zolamk: I was thinking you could get the latest reviews for a pull request using that query and filter out the ones that have approved, that way you won't need to sort and you'll also have everybody who reviewed the pull request right, what do you think?

shay2025: Yes, that seems to be a great solution 🙌

shay2025: @zolamk do you confirm that the response will only contain unique reviewers?

zolamk: Yes, it will only contain the latest reviews from each reviewer

zolamk: this is the sample data
```
{
  "data": {
    "repository": {
      "pullRequest": {
        "latestReviews": {
          "nodes": [
            {
              "id": "PRR_kwDOHEnsLs5PEh3p",
              "state": "CHANGES_REQUESTED",
              "author": {
                "login": "ferreiratiago"
              }
            },
            {
              "id": "PRR_kwDOHEnsLs5O6Dj_",
              "state": "APPROVED",
              "author": {
                "login": "shay2025"
              }
            },
            {
              "id": "PRR_kwDOHEnsLs5PInKd",
              "state": "APPROVED",
              "author": {
                "login": "reviewpad"
              }
            }
```

zolamk (changes requested):
Minor comment

zolamk (approved):
LGTM 👍

codehost/github/target/pull_request_target.go
```diff
@@ -539,3 +539,47 @@ func (target *PullRequestTarget) GetProjectV2ItemID(projectID string) (string, e
...
```
ferreiratiago, line 543: This naming is miss leading. It does not return the approvaded reviews but the latest approved reviews. Please rename it to `GetLatestApprovedReviews`.

utils/array.go
```diff
@@ -13,3 +13,13 @@ func ElementOf(s []string, str string) bool {
...
+
+func Contains(slice []string, str string) bool {
+	for _, elem := range slice {
+		if elem == str {
+			return true
+		}
+	}
+
+	return false
+}
```
ferreiratiago, line 25: This function looks the same as the one above. Please use `ElementOf`.

ferreiratiago (changes requested):

ferreiratiago (approved):

PR #824, opened 2023-04-08
feat: compact format for if, then and actions

zolamk:
## Description
Introduces a compact form for `if`, `then` and `actions` fields in reviewpad file.
```
workflows:
  - name: add-label
    if: '$size() <= 30'
    then: '$addLabel("small")'

pipelines:
  - name: renovate-bot
    trigger: $author() == "renovate[bot]"
    stages:
      - actions: $assignAssignees($group("owners"), 1)
        until: $assignees() != []
```

## Related issue

Closes reviewpad/raas-common#103

## Type of change

## How was this tested?
Manual and Unit tests.

## Checklist

**Ask**: this pull request requires a code review before merge


engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
 }
+
+func processCompactRules(rawRule any, currentRules []PadRule) ([]PadRule, []PadWorkflowRule, error) {
+	var rule *PadRule
+	var workflowRule *PadWorkflowRule
+	var rules []PadRule
+	var workflowRules []PadWorkflowRule
+
+	switch r := rawRule.(type) {
+	case string:
```
ferreiratiago, line 107: Can we add some comments for each case? If my understanding is correct:

```go
switch r := rawRule.(type) {
// Specificition of inline rule or rule name.
case string:
	// ...
// List of rules inline rules or rule names.
case []any:
	// ...
// Specificition of "rule" with/without "extra-actions".
case map[string]any:
	// ...
default:
	return nil, nil, fmt.Errorf("unknown rule type %T", r)
}
```

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
+		rule = decodeRule(r)
```
ferreiratiago, line 108: If my understanding is correct being string can mean the name of a specified rule under `rules`, which means that this decode is not correct. It works because on line [136](https://github.com/reviewpad/reviewpad/pull/824/files#diff-ea18868fc6cfa08e95023e74934900b2af69a653d53a65a962abe70616a132f6R136) we look for all rules and we will end up with the correct spec.

We don't need to correct this since it's working and we will be addressing this issue later.


Just wanted to create awareness.

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
+			return nil, nil, err
+		}
+
+		workflowRule = decodedWorkflowRule
+		rule = decodeRule(decodedWorkflowRule.Rule)
+	default:
+		return nil, nil, fmt.Errorf("unknown rule type %T", r)
+	}
+
+	// we need to check rule is not nil if we are processing a list of rules
```
ferreiratiago, line 134: No need for this comment.

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
+	if rule != nil {
+		if _, exists := findRule(currentRules, rule.Name); !exists {
+			rules = append(rules, *rule)
+		}
+	}
+
+	// we need to check workflowRule is not nil if we are processing a list of rules
```
ferreiratiago, line 141: No need for this comment.

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
 }
 
-func decodeWorkflowRule(rule map[string]interface{}) (*PadWorkflowRule, error) {
+func decodeWorkflowRule(rule map[string]any) (*PadWorkflowRule, error) {
 	workflowRule := &PadWorkflowRule{}
 	err := mapstructure.Decode(rule, workflowRule)
 	return workflowRule, err
 }
+
+func processCompactRules(rawRule any, currentRules []PadRule) ([]PadRule, []PadWorkflowRule, error) {
```
ferreiratiago, line 100: IMO this name is not the best since we are processing all rules regardless of being compact. I would suggest renaming to `normalizeRules`. Also, to add a comment to the function explaining what it does.

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
+
+	// we need to check workflowRule is not nil if we are processing a list of rules
+	if workflowRule != nil {
+		workflowRules = append(workflowRules, *workflowRule)
+	}
+
+	return rules, workflowRules, nil
+}
+
+func processCompactActions(nonNormalizedActions any) ([]string, error) {
```
ferreiratiago, line 149: The same here. I would suggest renaming to `normalizeActions` and to add a comment explaining what the function does.

ferreiratiago (changes requested):

engine/inline_rules_normalizer.go
```diff
@@ -99,8 +91,81 @@ func decodeRule(rule string) *PadRule {
...
+	case []any:
+		for _, rawAction := range action {
+			processedActions, err := processCompactActions(rawAction)
+			if err != nil {
+				return nil, err
+			}
+
+			actions = append(actions, processedActions...)
+		}
+	case nil:
```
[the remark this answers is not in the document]
zolamk: that's in case the workflow doesn't have any actions but only extra actions associated with the rules

engine/inline_rules_normalizer.go
zolamk: Yes, this is just creating a rule structure using the spec as a name and we verify it doesn't exist at a later stage, it was a way of avoiding having duplicate rules if the same spec is repated

ferreiratiago (approved):

PR #834, opened 2023-04-18
feat: process if then else block as a run

zolamk:
## Description
introduces an `else` block in reviewpad workflows, this pull request will enable arbitrarily nested conditionals in reviewpad `if ... then ... else` blocks in addition to introducing `else`.

### <samp>🤖 Generated by Copilot at 474de44</samp>

Added support for extra actions in rules and a new syntax for the top level `if ... then ... else` block in workflows. Modified the `PadWorkflow` type, the `processWorkflow` function, and the `getActionsFromRunBlocks` function to handle the new features.

## Related issue

Closes reviewpad/raas-common#102

## Type of change

## How was this tested?
Manual and Unit tests

## Checklist

**Ask**: this pull request requires a code review before merge

### <samp>🤖 Generated by Copilot at 474de44</samp>

*  Add `extraActions` parameter to `getActionsFromRunBlocks` function to execute additional actions before run actions ([link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-a265bb7df9128e988c6354a6c19295529954daecf10338751f3d394e0298185fL270-R270), [link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-a265bb7df9128e988c6354a6c19295529954daecf10338751f3d394e0298185fL277-R277), [link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-a265bb7df9128e988c6354a6c19295529954daecf10338751f3d394e0298185fL284-R284), [link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-a265bb7df9128e988c6354a6c19295529954daecf10338751f3d394e0298185fL292-R292))
*  Handle new syntax of workflow definition that allows top level `if ... then ... else` block as a run block ([link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-ea18868fc6cfa08e95023e74934900b2af69a653d53a65a962abe70616a132f6L70-R88), [link](https://github.com/reviewpad/reviewpad/pull/834/files?diff=unified&w=0#diff-75c510fa57c2a4141549710b101cacf6718f98bbcd315480f224178a824a3701R112))

### DIFF
--- a/engine/inline_rules_normalizer.go
+++ b/engine/inline_rules_normalizer.go
@@ -67,27 +67,25 @@ func processWorkflow(workflow PadWorkflow, currentRules []PadRule) (*PadWorkflow
 		On:          workflow.On,
 	}
 
-	actions, err := normalizeActions(workflow.NonNormalizedActions)
+	// since the top level if ... then ... else block can be interpreted as a run block
+	// we are converting it to a run block and then prepending it to the rest of the runs
+	workflowRun, workflowRules, err := normalizeRun(map[string]any{
+		"if":   workflow.NonNormalizedRules,
+		"then": workflow.NonNormalizedActions,
+		"else": workflow.NonNormalizedElse,
+	}, currentRules)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	wf.Actions = actions
-
-	compactRules, workflowRules, err := normalizeRules(workflow.NonNormalizedRules, currentRules)
+	runs, runRules, err := normalizeRun(workflow.NonNormalizedRun, append(currentRules, workflowRules...))
 	if err != nil {
 		return nil, nil, err
 	}
 
-	runs, runRules, err := normalizeRun(workflow.NonNormalizedRun, append(currentRules, compactRules...))
-	if err != nil {
-		return nil, nil, err
-	}
+	wf.Runs = append(workflowRun, runs...)
 
-	wf.Rules = workflowRules
-	wf.Runs = runs
-
-	return wf, append(compactRules, runRules...), nil
+	return wf, append(workflowRules, runRules...), nil
 }
 
 func decodeRule(rule string) *PadRule {
--- a/engine/inline_rules_normalizer.go
+++ b/engine/inline_rules_normalizer.go
@@ -67,25 +67,32 @@ func processWorkflow(workflow PadWorkflow, currentRules []PadRule) (*PadWorkflow
 		On:          workflow.On,
 	}
 
-	// since the top level if ... then ... else block can be interpreted as a run block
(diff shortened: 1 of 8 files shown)


engine/inline_rules_normalizer.go
```diff
@@ -67,27 +67,32 @@ func processWorkflow(workflow PadWorkflow, currentRules []PadRule) (*PadWorkflow
...
 
-	wf.Actions = actions
-
-	compactRules, workflowRules, err := normalizeRules(workflow.NonNormalizedRules, currentRules)
-	if err != nil {
-		return nil, nil, err
-	}
+	// process only if the workflow has an if ... then ... else block
+	// since there might be cases where the workflow only has a run block
+	if workflow.NonNormalizedRules != nil {
```
ferreiratiago, line 77: Can we say that if `workflow.NonNormalizedRun` is null then we assume that we have an `if/then/else` block and convert it to a `run` block before `normalizeRun` on line [70](https://github.com/reviewpad/reviewpad/pull/834/files#diff-ea18868fc6cfa08e95023e74934900b2af69a653d53a65a962abe70616a132f6R70).

Something similar to this:

```go
func processWorkflow(workflow PadWorkflow, currentRules []PadRule) (*PadWorkflow, []PadRule, error) {
	wf := &PadWorkflow{
		Name:        workflow.Name,
		Description: workflow.Description,
		AlwaysRun:   workflow.AlwaysRun,
		Rules:       workflow.Rules,
		Actions:     workflow.Actions,
		On:          workflow.On,
	}

	// If there is no run, then we assume that the workflow is a if/then/else block.
	// In this case we need to convert the if/then/else block to a run block.
	if workflow.NonNormalizedRun == nil {
		workflow.NonNormalizedRun = map[string]any{
			"if":   workflow.NonNormalizedRules,
			"then": workflow.NonNormalizedActions,
			"else": workflow.NonNormalizedElse,
		}
	}

	runs, runRules, err := normalizeRun(workflow.NonNormalizedRun, currentRules)
	if err != nil {
		return nil, nil, err
	}

	wf.Runs = runs

	return wf, runRules, nil
}
```

WDYT?

ferreiratiago (changes requested):

ferreiratiago (changes requested):
Unfortunately, the following test case is not working properly.
```
workflows:
  - name: Test if with only extra action
    always-run: true
    if:
      - rule: tautology
        extra-actions:
          - $info("✅ OK")
```

ferreiratiago (approved):

